PR #248, opened 2012-11-06
ECMS-4285: [Content Explorer] Column "Document Title" instead of "Name"

cuonglh:
Fix description: Change to display title key instead of name key


trangvh:
Won't fix in Platform 3.5.x.

PR #703, opened 2014-09-02
ECMS-6170 | Dear null is shown in notification mail when watching a docu...

tmarwen:
...ment
* Problem Analysis:
When using the ```getDisplayName()``` it may return the ```displayName``` whithout checking its nullity when it is delegated to ```org.exoplatform.services.organization.idm.UserImpl```.
* Fix Description:
The current version of ```EmailNotifyListener``` must use the method ```getFullName()``` method even if it is deprecated until the changes in GTNPORTAL-3482 are applied.


trangvh:
Merged to stable/4.1.x.

PR #708, opened 2014-09-23
ECMS-6401 | Cursor not displayed as hand when try to choose folder path ...

tmarwen:
...or content in CLV
**Fix Description:**
* Add css pointer style to file list items.


trangvh:
Replace by #801

PR #725, opened 2014-10-09
ECMS-5834:Missing resource bundle keys for some tooltips

vannd:
Fix Description:  
- Add resource bundle to webui_en.xml
- ECMS-5095: Add new key for Permission Metadata management


trangvh:
Merged into stable/4.1.x.

PR #740, opened 2014-10-27
ECMS-6462: [ECMAdmin][Add Drive] Failed to select Node types

trangvh:
Fix description:
* Restore the usage of UIFormCheckBoxInput in UINodeTypeSelector.


trangvh:
Refused. Use #786 instead.

PR #742, opened 2014-10-29
ECMS-6445: [File Search] Display document of frozenNode in search result...

ttthanh:
...s

Fix description:
- Filter result during pagination progress


trangvh:
Merged to stable/4.1.x.

PR #746, opened 2014-11-11
ECMS-6479: Cannot cut and paste a document into a Content Folder

vannd:
Fix Description:

* Check clipboard when calling getClipboardList. Clean commands in clipboard relate to a node which was deleted


vannd:
Merge to stable/4.1.x.

PR #748, opened 2014-11-11
ECMS-6272: [Simple Search] Duplicate results when keyword matches file c...

ttthanh:
...ontent

Problem analysis
- With ECMS-5758, the nt:resource file is indexed. Each nt:file has a nt:resource as a child to store content. Moreover nt:resource nodetype is one of instance of nt:base. As the result, in the search result, user can see both of nt:file and its nt:resource.

Fix description:
- Exclude nt:resource from search results


trangvh:
Merged to stable/4.1.x.

PR #749, opened 2014-11-11
Fix/4.1.x/ecms 6460

ttthanh:


vannd:
Merge to stable/4.1.x

PR #755, opened 2014-11-13
ECMS-6415: [Content Selector] "Web Content" document in the left tree fo...

vannd:
...r browsing

Fix description:
- Use 'type' parameter to check the context: if type = 'editor' display webcontent/illustrated webcontent in left tree, otherwise they will be invisible


vannd:
Merge to stable/4.1.x

PR #758, opened 2014-11-20
ECMS-6474: [Webdav - Windows] NPE and redundant comment when uploading a...

ttthanh:
... file via webdav of windows

Problem analysis
- FILE_CREATED_ACTIVITY event is fired when updating node in webdav
- FILE_EDIT_ACTIVITY event is fired when uploading file in windows webdav. This behavior is caused by specification of this type of webdav. The PUT action is called twice to create empty file then update content.

Fix description
- Only fire FILE_CREATED_ACTIVITY event when uploading file.
- Do not fire FILE_EDIT_ACTIVITY event when uploading file by webdav in windows.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/webdav/WebDavServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/webdav/WebDavServiceImpl.java
@@ -436,9 +436,7 @@ public class WebDavServiceImpl extends org.exoplatform.services.jcr.webdav.WebDa
         session = item.getSession();
       }
       activityService = WCMCoreUtils.getService(ActivityCommonService.class);
-      try {
-        currentNode = (Node) session.getItem(path(repoPath));
-      } catch (Exception e) {
+      if (!session.itemExists(path(repoPath))) {
         isCreating = true;
       }
     } catch (PathNotFoundException exc) {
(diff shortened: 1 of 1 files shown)


core/services/src/main/java/org/exoplatform/services/cms/webdav/WebDavServiceImpl.java
```diff
@@ -433,6 +435,12 @@ public Response put(@PathParam("repoName") String repoName,
             + LinkUtils.createPath(item.getPath(), Text.escapeIllegalJcrChars(LinkUtils.getItemName(path(repoPath))));
         session = item.getSession();
       }
+      activityService = WCMCoreUtils.getService(ActivityCommonService.class);
+      try {
+        currentNode = (Node) session.getItem(path(repoPath));
+      } catch (Exception e) {
```
anhvurz90: catch(Exception) -> Not OK!

core/services/src/main/java/org/exoplatform/services/cms/webdav/WebDavServiceImpl.java
```diff
@@ -433,6 +435,12 @@ public Response put(@PathParam("repoName") String repoName,
             + LinkUtils.createPath(item.getPath(), Text.escapeIllegalJcrChars(LinkUtils.getItemName(path(repoPath))));
         session = item.getSession();
       }
+      activityService = WCMCoreUtils.getService(ActivityCommonService.class);
+      try {
+        currentNode = (Node) session.getItem(path(repoPath));
```
anhvurz90: Should use session.itemExists()

PR #780, opened 2015-01-13
ECMS-6298: [WebDAV] Check file size before uploading

trangvh:
Fix description:
* Use the new file-size-limit of JCR WebDavService's init parameters.
* Convert InputStream to ByteArrayInputStream to get its size without changing the initial stream.
* Upload file only it is smaller than the size threshold. Otherwise return 413 "Entity is too large".


trangvh:
Close as Won't fix.

PR #782, opened 2015-01-20
ECMS-6597: Improve UI message when importing xml document contain invali...

vannd:
...d charater

Fix description: Get position information of invalid character in JCR exception and show on UI message


vannd:
Merge to stable/4.1.x.

PR #785, opened 2015-01-26
ECMS-6185:SuperType is mandatory while adding a mixin

vannd:
Fix Description: Add Onchange event to MixinType select box to remove MandatoryValidator of SuperType if MixinType is true


trangvh:
Merged to develop.

PR #786, opened 2015-01-28
ECMS-6462:[Content Administrator] [Add Drive] Failed to select Node Type...

vannd:
...s

Fix description: use getName instead of getValue of UICheckBoxInput object


vannd:
Merged  to stable/4.1.x.

PR #787, opened 2015-01-29
ECMS-6561: Too many logs when previewing documents

trangvh:
Fix description:
* Only LOG in debug mode. Remove unused imports


trangvh:
Merged to stable/4.1.x.

PR #788, opened 2015-01-30
ECMS-6423 [PDF Viewer] Missing file information when there isn't preview

trangvh:
Fix description: Give more information of the file in 2 cases:
* The document has more than 100 pages
* Office file is larger than 5 MB


trangvh:
Merged to stable/4.1.x.

PR #790, opened 2015-02-02
ECMS-6659: Use ConversationState instead of OrganizationService if possi...

trungthanh-tran:
...ble

Fix description:
- Replace several functions of OrganizationService by using ConversationState

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/queries/impl/QueryServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/queries/impl/QueryServiceImpl.java
@@ -571,6 +571,9 @@ public class QueryServiceImpl implements QueryService, Startable{
    * @return
    */
   private boolean hasMembership(String userId, String roleExpression) {
+    if (userId == null || userId.length() == 0) {
+      return false;
+    }
     if(roleExpression.equals("*") || roleExpression.equals(IdentityConstants.ANY))
       return true;
     ConversationState conversationState = ConversationState.getCurrent();
--- a/core/services/src/main/java/org/exoplatform/services/cms/queries/impl/QueryServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/queries/impl/QueryServiceImpl.java
@@ -35,14 +35,18 @@ import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.services.organization.MembershipHandler;
 import org.exoplatform.services.organization.OrganizationService;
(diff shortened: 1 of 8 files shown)


core/services/src/main/java/org/exoplatform/services/cms/queries/impl/QueryServiceImpl.java
```diff
@@ -569,16 +573,26 @@ private boolean canUseQuery(String userId, Node queryNode) throws Exception{
...
+    Identity identity = conversationState.getIdentity();
     String membershipType = roleExpression.substring(0, roleExpression.indexOf(":"));
     String groupName = roleExpression.substring(roleExpression.indexOf(":") + 1);
     try {
       MembershipHandler membershipHandler = organizationService_.getMembershipHandler();
       if ("*".equals(membershipType)) {
-        // Determine if there exists at least one membership
-        return !membershipHandler.findMembershipsByUserAndGroup( userId,groupName).isEmpty();
+    	// Determine if there exists at least one membership
+        if (userId.equals(ConversationState.getCurrent().getIdentity().getUserId())) {
```
trangvh: should check userId == null before that line.

core/services/src/main/java/org/exoplatform/services/wcm/utils/WCMCoreUtils.java
```diff
@@ -183,7 +188,22 @@ public static boolean hasPermission(String userId, List<String> permissions, boo
     try {
       OrganizationService organizationService = WCMCoreUtils.getService(OrganizationService.class);
       startRequest(organizationService);
-      Collection<?> memberships = organizationService.getMembershipHandler().findMembershipsByUser(userId);
+      Identity identity = ConversationState.getCurrent().getIdentity();
+      Collection<?> memberships = null;
+      if (userId.equals(identity.getUserId())){
```
trangvh: should handle the case (userId == null) before.

ext/authoring/webui/src/main/java/org/exoplatform/wcm/authoring/listener/PostUpdateStateEventListener.java
```diff
@@ -136,8 +137,13 @@ private void sendMail(Node node,
 
       ListAccess<User> userList = userh.findUsersByGroupId(group);
       User currentUser = null;
+      ConversationState conversationState = ConversationState.getCurrent();
       try {
-        currentUser = userh.findUserByName(userId);
+        if (userId.equals(conversationState.getIdentity().getUserId())) {
```
trangvh: check if (userId == null)

PR #795, opened 2015-02-17
ECMS-6660:[PDF Viewer] Persistent PDF files in temp folder

AtiaZ:


AtiaZ:
In order to avoid the using of the method **deleteOnExit()** which is deleted after this comment 
```
/*
       file.deleteOnExit();
         PM Comment : I removed this line because each deleteOnExit creates a reference in the JVM for future removal
         Each JVM reference takes 1KB of system memory and leads to a memleak
       */
```
After much research and several comparison between the different methods, we noticed that the method can solve the problem with less memory and CPU consumption.

```
  final File finalCont = file;
       Runtime.getRuntime().addShutdownHook(new Thread() {
         public void run() {
           finalCont.delete();
         }
       });
```
This method helps to delete temp files when the Platform server stops.

PR #820, opened 2015-05-26
ECMS-6902: [IE9, IE10] Document Preview doesn't work

trangvh:
Problem analysis:
* It lacks some JavaScript supports in IE 9 and IE 10 browsers.
  pdf.js handles this in compatibility.js but this file isn't loaded in Platform.

Fix description:
* Load compatibility.js in PDFJSViewer.gtmpl.


trangvh:
Merged into Platform 4.2.0.

PR #828, opened 2015-06-03
ECMS-6935: [IE11][WCM Content Selector] Incorrect display when upload…

vannd:
…ing a file

Fix description:
* Hide Upload button after selecting upload a file


trangvh:
Merged to PLF 4.2.1.

PR #842, opened 2015-06-26
ECMS-6474 fix possible NPE when uploading document via mobile app

rdenarie:


trangvh:
Merged to Platform 4.2.1.

PR #850, opened 2015-09-15
Fix/ecms 6992

chautn:


chautn:
replaced by https://github.com/exodev/ecms/pull/143

PR #852, opened 2015-09-16
ECMS-7128 [Open in Office] Unable to save a document to Public folder

dongpd:
Analysis:
- Office file has thubmnail view.
- Updating an office file cause removing its old thumbnail.
- Login user can have no permission to edit on thumbnail folder
  Solution:
- Thubnail folder is specific for the system then we should always use system session to remove thumbmail

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/wcm/utils/WCMCoreUtils.java
+++ b/core/services/src/main/java/org/exoplatform/services/wcm/utils/WCMCoreUtils.java
@@ -716,4 +716,11 @@ public class WCMCoreUtils {
     }
     return barNavigationStyle;
   }
+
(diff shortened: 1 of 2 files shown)


core/services/src/main/java/org/exoplatform/services/wcm/utils/WCMCoreUtils.java
```diff
@@ -716,4 +716,11 @@ public static String getBarNavigationStyle() {
     }
     return barNavigationStyle;
   }
+
+  public static Node getNodeBySystemSession(Node node) throws RepositoryException {
+    node.getSession().save();
```
minhdv: Why do you need to save the node before get it by system provider?

PR #864, opened 2015-12-01
ECMS-7271 Wrong display of PDF content after modification.

dongpd:
Analysis: jcr:lastModified of jcr:content is not updated after changing jcr:data
Solution: add condition comparing jcr:data stream size with file size to regenerate pdf and update pdfcache


trangvh:
The patch was delivered to the customer.

PR #865, opened 2015-12-02
[ECMS-6069] Send many emails when user watch a edited document

AtiaZ:
fix description : 
Send just one email when clicking on "Save & Close" or on "Save" after making changes in the document metadata.


core/services/src/main/java/org/exoplatform/services/cms/watch/impl/EmailNotifyListener.java
```diff
@@ -92,7 +92,8 @@ public void onEvent(EventIterator arg0) {
     WatchDocumentServiceImpl watchService = (WatchDocumentServiceImpl)WCMCoreUtils.getService(WatchDocumentService.class);
     MessageConfig messageConfig = watchService.getMessageConfig();
     List<String> emailList = getEmailList(NodeLocation.getNodeByLocation(observedNode_));
-    for (String receiver : emailList) {
+    if (arg0.getSize() > 5) {
```
thomasdelhomenie, line 95: Why 5 ?
It should probably be set in a constant, and add some comments.

thomasdelhomenie (changes requested):

boubaker (changes requested):
Please make the send email async.

AtiaZ:
Replaced by https://github.com/exodev/ecms/pull/264

PR #868, opened 2015-12-08
ECMS-7254: add treatments for metadata property of type Double

wkhessairi:


core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/popup/info/UIViewMetadataForm.java
```diff
@@ -173,6 +177,19 @@ public void execute(Event<UIViewMetadataForm> event) throws Exception {
               }
               if(!node.hasProperty(name) || (node.hasProperty(name) && !node.getProperty(name).getString().equals(value)))
                 node.setProperty(name, value);
+            } else if (requiredType == 4){ // double
+              UIFormInput uiInput = uiForm.getUIInput(inputName);
+              double value = 0;
+              if(uiForm.getUIInput(inputName) != null) {
+                if ((!node.hasProperty(name) || (node.hasProperty(name) && node.getProperty(name).getDouble() != value))) {
```
boubaker: Could you plz add explicitly != 0 instead of != value ?

core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/popup/info/UIViewMetadataForm.java
```diff
@@ -173,6 +177,19 @@ public void execute(Event<UIViewMetadataForm> event) throws Exception {
               }
               if(!node.hasProperty(name) || (node.hasProperty(name) && !node.getProperty(name).getString().equals(value)))
                 node.setProperty(name, value);
+            } else if (requiredType == 4){ // double
+              UIFormInput uiInput = uiForm.getUIInput(inputName);
+              double value = 0;
+              if(uiForm.getUIInput(inputName) != null) {
```
boubaker: Plz use variable uiInput instead of "uiForm.getUIInput(inputName)" and you should also test on emptiness of value of the field ( StringUtils.isNotBlank(uiInput.getValue()) )

wkhessairi: > you should also test on emptiness of value of the field ( StringUtils.isNotBlank(uiInput.getValue()) )

I think that we should let user choose if he can save an empty value, otherwise this field will be considered as required, and we oblige user to set a non empty value

core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/popup/info/UIViewMetadataForm.java
```diff
@@ -173,6 +173,19 @@ public void execute(Event<UIViewMetadataForm> event) throws Exception {
...
                 node.setProperty(name, value);
+            } else if (requiredType == 4){ // double
+              UIFormInput uiInput = uiForm.getUIInput(inputName);
+              double value = 0;
+              if(uiInput != null && StringUtils.isNotBlank((String)uiInput.getValue())) {
+                if ((!node.hasProperty(name) || (node.hasProperty(name) && node.getProperty(name).getDouble() != 0))) {
+                  try {
+                    value =  Double.parseDouble((String) uiInput.getValue());
+                    node.setProperty(name, value);
+                  } catch (Exception e) {
```
boubaker: You should catch only the exception "NumberFormatException" and you shouldn't delete the value, but you should discard changes and inform the user that the number format is wrong for this field. So add a LOG.error + throw a new Runtime Exception with an internationalized message for this.

core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/popup/info/UIViewMetadataForm.java
```diff
@@ -173,6 +173,19 @@ public void execute(Event<UIViewMetadataForm> event) throws Exception {
               }
               if(!node.hasProperty(name) || (node.hasProperty(name) && !node.getProperty(name).getString().equals(value)))
                 node.setProperty(name, value);
+            } else if (requiredType == 4){ // double
+              UIFormInput uiInput = uiForm.getUIInput(inputName);
+              double value = 0;
+              if(uiInput != null && StringUtils.isNotBlank((String)uiInput.getValue())) {
+                if ((!node.hasProperty(name) || (node.hasProperty(name) && node.getProperty(name).getDouble() != 0))) {
```
boubaker: Add an else block where you delete the previous property if the input is null.

boubaker (changes requested):

wkhessairi:
changes done

PR #870, opened 2016-01-05
ECMS-7381: Errors when creating new portal of Empty/ACME template

trangvh:
Fix description:
- Replace {portalName} by siteName from input.


mgreau:
Moved to exodev#173

PR #874, opened 2016-03-28
PLF-6566: [Fuzzy Query] Harmonize the use of property «exo.unified-se…

kmenzli:
…arch.engine.fuzzy.enable»


boubaker (changes requested):
Could you add unit tests here plz and recreate the PR on exodev repository.

PR #876, opened 2016-05-13
ECMS-7430: Change the ping frequency according to the server reply

vsellier:
PR content :
- Use setTimeout instead of setInterval to change the call frequency easily
- Slow down the calls if the response code is the 500 range
- Stop the calls for all the other errors


vsellier:
Wrong base, replaced by https://github.com/exodev/ecms/pull/194

PR #877, opened 2016-05-27
ECMS-7444:Property definition is not found exception after deploying …

mkrout:
…the "Spaces-Categories" Add-on


thomasdelhomenie:
LGTM
@boubaker can you have a look please (since it is related to the clean up of Social JCR nodes) ?

mkrout:
New PR https://github.com/exodev/ecms/pull/202

boubaker:
Just some questions to know if this is the best solution:
- Shouldn't we use collaboration WS instead of Social for ECMS categories ? (So the new addon will not face this problem anymore)
- Why do we need exo:dateModified property ? In the source code, I don't see why we should add a persistent information while the usage is for processing (to know if the node is new or not).

mkrout:
- We are using the collaboration WS to store categories, but the issue is related to the target node of the category ("soc:spacedefinition") that is stored in social WS.
- I don't want to delete this property since I don't know if it is needed somewhere or not.

PR #878, opened 2016-09-05
ECMS-7454

azaoui:
ECMS-7454:should not use ThreadLocal for user navigation

### DIFF
--- a/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
+++ b/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
@@ -681,6 +681,29 @@
     </component-plugin>
   </external-component-plugins>
   
+  <!-- Configure User Navigation caches - Standalone mode -->
+  <external-component-plugins>
+    <target-component>org.exoplatform.services.cache.CacheService</target-component>
+    <component-plugin>
+      <name>addExoCacheConfig</name>
+      <set-method>addExoCacheConfig</set-method>
+      <type>org.exoplatform.services.cache.ExoCacheConfigPlugin</type>
+      <description>Configures the cache for User Navigation</description>
+      <init-params>
+        <object-param>
+          <name>CacheUserNavigationKeeper</name>
+          <description></description>
+          <object type="org.exoplatform.services.cache.ExoCacheConfig">
+            <field name="name"><string>CacheUserNavigationKeeper</string></field>
+            <field name="maxSize"><int>${wcm.cache.userNavigation.capacity:300}</int></field>
+            <field name="liveTime"><long>${wcm.cache.userNavigation.timetolive:600000}</long></field>
+            <field name="implementation"><string>org.exoplatform.services.cache.concurrent.ConcurrentFIFOExoCache</string></field>
+          </object>
+        </object-param>
+      </init-params>
+    </component-plugin>
+  </external-component-plugins>
+ 
   <!--Configure Lock service cached-->
   <external-component-plugins>
     <target-component>org.exoplatform.services.cache.CacheService</target-component>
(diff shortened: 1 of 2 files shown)


core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -681,6 +681,53 @@
...
+      <set-method>addExoCacheConfig</set-method>
+      <type>org.exoplatform.services.cache.ExoCacheConfigPlugin</type>
+      <description>Configures the cache for User Navigation</description>
+      <init-params>
+        <object-param>
+          <name>CacheUserNavigationKeeper</name>
+          <description></description>
+          <object type="org.exoplatform.services.cache.ExoCacheConfig">
+            <field name="name"><string>CacheUserNavigationKeeper</string></field>
+            <field name="maxSize"><int>${wcm.cache.userNavigation.capacity:300}</int></field>
```
ahamdi: We may need to set a bigger value for the cache capacity since we will have at least an entry per connected user. I suggest to have 5000 entries

core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -681,6 +681,53 @@
...
+            <field name="liveTime"><long>${wcm.cache.userNavigation.timetolive:600000}</long></field>
```
ahamdi: The time to live could be equal to the HTTP session, 30 mn = 30 \* 60 \* 1000 = 1800000 ms

core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -681,6 +681,53 @@
...
+          </object>
+        </object-param>
+      </init-params>
+    </component-plugin>
+  </external-component-plugins>
+
+    <!-- Configure User Navigation caches - Cluster mode -->
+  <external-component-plugins>
+    <target-component>org.exoplatform.services.cache.CacheService</target-component>
+    <component-plugin  profiles="cluster">
```
ahamdi: IMO This cache does not need to be shared in cluster

azaoui:
done

boubaker:
Two major remarks:
1/ This method is used to add in a javascript variable the list of navigations( [see here](https://github.com/exoplatform/ecms/blob/develop/packaging/wcm/webapp/src/main/webapp/groovy/SingleContentViewer/UISingleContentView.gtmpl#L14) ). This javascript variable is used only in this [dialog.gtmpl](https://github.com/exoplatform/platform/blob/develop/samples/wai-template/src/main/webapp/WEB-INF/conf/sample-portal/wcm/artifacts/nodetype-templates/siteBreadcrumbWebContent/dialogs/dialog1.gtmpl#L279)
So the method in UISingleContentView.gtmpl should be moved to the dialog1.gtmpl that use it.
2/ I 'm ok that we can use the ConversationState. But just FYI, it will be the same thing (with the following modification).
Why?

When a user requests a page containing an SCV, the "navigations" object is deleted at the beginning of the request ( [see here](https://github.com/exoplatform/ecms/blob/develop/packaging/wcm/webapp/src/main/webapp/groovy/portal/webui/workspace/UIWorkingWorkspace.gtmpl#L20) ) and it will be cached when the SCVis rendered ([see here](https://github.com/exoplatform/ecms/blob/develop/packaging/wcm/webapp/src/main/webapp/groovy/SingleContentViewer/UISingleContentView.gtmpl#L14)) (should be in dialog.gtmpl instead as mentioned in point 1/). (So it will not be destroyed at the end of the HTTP request).
Instead of doing this, the navigation should be cached at the first display of the dialog that needs it [here](https://github.com/exoplatform/platform/blob/develop/samples/wai-template/src/main/webapp/WEB-INF/conf/sample-portal/wcm/artifacts/nodetype-templates/siteBreadcrumbWebContent/dialogs/dialog1.gtmpl#L279) ) and should be destroyed when the page render is finished, [here](https://github.com/exoplatform/ecms/blob/develop/packaging/wcm/webapp/src/main/webapp/groovy/portal/webui/workspace/UIWorkingWorkspace.gtmpl#L49) after  renderChildren method. Attention to the case when an exception is raised, so it should be in a finally block.

PR #879, opened 2016-09-06
Fix/ecms 7454

azaoui:
ECMS-7454:should not use ThreadLocal for user navigation

### DIFF
--- a/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
+++ b/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
@@ -712,6 +712,30 @@
       </init-params>
     </component-plugin>
   </external-component-plugins>
+
+    <!-- Configure User Navigation caches - Cluster mode -->
+  <external-component-plugins>
+    <target-component>org.exoplatform.services.cache.CacheService</target-component>
+    <component-plugin  profiles="cluster">
+      <name>addExoCacheConfig</name>
+      <set-method>addExoCacheConfig</set-method>
+      <type>org.exoplatform.services.cache.ExoCacheConfigPlugin</type>
+      <description>Configures the cache for User Navigation</description>
+      <init-params>
+        <object-param>
+          <name>CacheUserNavigationKeeper</name>
+          <description></description>
+          <object type="org.exoplatform.services.cache.impl.infinispan.generic.GenericExoCacheConfig">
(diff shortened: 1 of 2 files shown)


core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -690,6 +690,53 @@
...
+      <set-method>addExoCacheConfig</set-method>
+      <type>org.exoplatform.services.cache.ExoCacheConfigPlugin</type>
+      <description>Configures the cache for User Navigation</description>
+      <init-params>
+        <object-param>
+          <name>CacheUserNavigationKeeper</name>
+          <description></description>
+          <object type="org.exoplatform.services.cache.ExoCacheConfig">
+            <field name="name"><string>CacheUserNavigationKeeper</string></field>
+            <field name="maxSize"><int>${wcm.cache.userNavigation.capacity:300}</int></field>
```
ahamdi, line 707: We may need to set a bigger value for the cache capacity since we will have at least an entry per connected user. I suggest to have 5000 entries

core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -690,6 +690,53 @@
...
+            <field name="liveTime"><long>${wcm.cache.userNavigation.timetolive:600000}</long></field>
```
ahamdi, line 708: The time to live could be equal to the HTTP session, 30 mn = 30_60_1000 = 1800000 ms

core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-services-configuration.xml
```diff
@@ -690,6 +690,53 @@
...
+            <field name="implementation"><string>org.exoplatform.services.cache.concurrent.ConcurrentFIFOExoCache</string></field>
+          </object>
+        </object-param>
+      </init-params>
+    </component-plugin>
+  </external-component-plugins>
+
+    <!-- Configure User Navigation caches - Cluster mode -->
+  <external-component-plugins>
+    <target-component>org.exoplatform.services.cache.CacheService</target-component>
```
ahamdi, line 718: IMO This cache does not need to be shared in cluster

PR #883, opened 2017-03-15
ECMS-7563 MissingResourceException: Cant find resource for bundle org.gatein.common.i18n.ParentChildResourceBundle, key Drives.label.Collaboration

nttuyen:
- Add missing resource bundle
- Fix display driver name as {0} Document


nttuyen:
Replaced by https://github.com/exodev/ecms/pull/276

PR #888, opened 2018-03-15
PLF-7913: can not open a file from search result

anoissi:


core/search/src/main/java/org/exoplatform/services/wcm/search/connector/DocumentSearchServiceConnector.java
```diff
@@ -144,7 +144,7 @@ protected String getPreviewUrl(ResultNode node, SearchContext context) throws Ex
       authorUsername = node.getProperty("exo:owner").getString();
     }
 
-    StringBuilder url = new StringBuilder("javascript:require(['SHARED/social-ui-activity'], function(activity) {activity.previewDoc(event, {doc:{");
+    StringBuilder url = new StringBuilder("javascript:require(['SHARED/social-ui-activity'], function(activity) {activity.previewDoc(null, {doc:{");
```
thomasdelhomenie, line 147: The event variable has been added in PLF-6975 (https://github.com/exoplatform/ecms/commit/098cc6992461815a18d08003af240bfb78e9c913#diff-c31f68308c48ad1e89da950bfcefbdb0) to fix an issue in mobile mode. Did you check PLF-6975 is still fixed with this change ?
We should find a way to pass the event instead (maybe by changing the href in onclick ?).

thomasdelhomenie (changes requested):

core/search/src/main/java/org/exoplatform/services/wcm/search/connector/DocumentSearchServiceConnector.java
anoissi: i have checked PLF-6975 with this change and it's Ok 
also i have changed href to onclick and it's work for this issue -> should i changes this pr as you suggested ? (href -> onclick)

thomasdelhomenie: > i have checked PLF-6975 with this change and it's Ok

So, is this parameter (event) useless ? If so we shoud remove it. Otherwise it is ok for me.

anoissi: Yes i have tested with this pr the issue PLF-6975 and it's work correctly

thomasdelhomenie: ok so you can remove it.

anoissi: Should i remove the parameter event from the function "previewDoc" ?, in fact this function is used by other scripts  (UiSharedContent.gtmpl...)

thomasdelhomenie: The parameter has been added in PLF-6975, so if PLF-6975 is still ok by removing it I guess we can remove it.
You mean it requires a lot of work ?

anoissi: When assign the value null to the parameter event DocumentSearchServiceConnector.java#getPreviewUrl and FileSearchServiceConnector.java#getPreviewUrl there is no problem and the issue PLF-6975 still fixed but there is other classes that are using this function previewDoc (event,settings)(UiSharedContent.gtmpl....)  ==> if i remove the parameter event (so previewDoc (event,settings)-> previewDoc (settings)) what test(other than PLF-6975) should i do to ensure that there is no regression

thomasdelhomenie: Let's keep things like that to keep minimal changes and avoid any risk.

thomasdelhomenie (approved):

PR #897, opened 2019-05-10
PLF-8175 : Display problem for arabic in the document thumbnail in Activity stream

AzmiTouil:


boubaker:
replaced by https://github.com/exodev/ecms/pull/470

PR #933, opened 2019-11-28
ECMS-7993: Interruption in ChangeStateCronJob at first corrupted node…

mzorai:
… (#502)

* ECMS-7993: Interruption in ChangeStateCronJob at first corrupted node


ahamdi (approved):
LGTM

PR #934, opened 2019-11-29
Fix/PLF-8622: Make sure that method createPageList doesn't throw an error.

mdiha:
When clicking on show more results after accessing the page list for unified search results an error is thrown. After digging in this issue i found out that a sublist method is executed while the second parameter can be of size zero. I added a check to avoid executing the sublist method if the parameter is of size zero.


ahamdi (changes requested):
We need to find what caused the issue in 5.1 and not in 5.3/6.0

mdiha:
Issue reproduced on 6.x.x, i close this PR.

PR #935, opened 2019-12-05
Fix/PLF-8622: Make sure that Unified Search doesn't throw an  ERROR in the logs while clicking on "Show more results" (#504)

mdiha:
While digging in this issue we found out that a process of subList method throws the error when there is no results. We added a condition for the case where the results size is lower than the offset to avoid executing the subList memthod.


mzorai (approved):
LGTM

PR #956, opened 2020-02-26
Fix/ECMS-7935: Getting the activity id of searched file

itouir:
* Getting the activity id of searched file
* Indexing the activity id


ahamdi (approved):
LGTM

PR #990, opened 2020-07-23
MAINT-26897 Fix Activity composer performances

boubaker:
Activity Composer is instanciated multiple times in Activity Stream: one for the activity composer to post new activities and other instances, one per Activity present in Stream to allow editing activity in composer. this strategy is very heavy on browser especially when posting new activity, in fact the AS gets updated and the Composer gets instantiated again each time there is an update in the AS (like, comment...)
To avoid this performance regression, a single instance of Activity composer will be used, and events through document.dispatchEvent will be used to open composer with editing activity details. See https://github.com/Meeds-io/social/pull/220


mkrout (approved):
LGTM

PR #1005, opened 2020-09-04
MAINT-27305: Make sure that no error appears when pasting a folder.

mdiha:
The issue is that when pasting a folder an error pop up shows up, but the folder is correctly pasted.
After digging in this issue i found out that in the process of pasting the folder there is an action of deleting a thumbnail node which is null so i added a check to avoid this error.


ahamdi (approved):

ahamdi (dismissed):
wrong repo, please recreate PR

mdiha:
> wrong repo, please recreate PR

Couldn't find the right repo.

hbenali:
here https://github.com/exodev/ecms

mdiha:
> here https://github.com/exodev/ecms

Thanks

PR #1010, opened 2020-09-11
Optimize indexed file contents in Elasticsearch (#704)

Jihed525:
Currently content of all file types are indexed in ES while, it's whorthless to index content of files that aren't readable like images, videos, zip files...
This improvement will add two properties in Files Indexing Connector to limit file contents to index by:
1/ Mimetype
2/ File content size
When the file content or mimetype doesn't match allowed both conditions, an empty content of file will be sent to ES.

(cherry picked from commit ac51926dca1c2e1dc149569fbf34945c8d17cbf1)


ahamdi (approved):
LGTM
@Jihed525 Documentation is required, you need to add those properties to the configuration section of eXo platform

PR #1022, opened 2020-10-05
MAINT-28623 : Create file activity just when commented (#711)

ahamdi:
This fix makes sure that :

 - File activities are created when the files:spaces activity type and the comment activity type are enabled
 - For every action we have a comment activity type that we can enable/disable
 - Only when commenting directly a File will generate the activity will be created with a comment

(cherry picked from commit 7510b8ede128a3fadcd0a1c70824389a9396e321)


ahmadmido (approved):
LGTM

PR #1030, opened 2020-10-12
 MAINT-28315 : unify how we get the mapping of extensions + mimetypes…

ahamdi:
Mimetypes are checked in ECMS and Kernel differently as they use two different list of extensions + mimetypes.
The fix does :
 - Unify the content of both lists
 - Make it possible to provide an external common file for mimetype properties to be used by mimetype resolvers

(cherry picked from commit 110c1feabe6b17160f94d3df7efc6bc91dd74ae3)


ahmadmido (approved):
LGTM

PR #1031, opened 2020-10-13
MAINT-28584 : display upload button after startup

ahamdi:
Upload button got escaped since there is another button that starts with Upload text (Upload New version)
This fix uses a better way to check the button availability and filters Cloud Drive buttons.
(cherry picked from commit 3c15b7b52f3dfdffdcd6d6cb4244778c3de24305)


azayati (approved):
LGTM

PR #1057, opened 2020-12-21
Fix view and open file icons UI

wafakd:
Document Search Result: UI bug when click open or view file on Documents table


ahamdi (changes requested):
Could you remove the console.log

ahamdi (approved):
LGTM

PR #1061, opened 2020-12-25
TASK-37045: Update new created document language before opening it in onlyOffice

aelbenney:
After modifying the three templates languages settings in [only office](https://github.com/exoplatform/onlyoffice/pull/71), We need to modify each document editing language to the user's platform one.


ahamdi (approved):
LGTM

PR #1064, opened 2020-12-29
37648: Update icon of application drive

ahamdi:


hbenali (commented):
This PR is already merged on the develop branch, it will be available after the next rebase job

PR #1072, opened 2021-01-10
MAINT-39276 : Fix ecms notifications plugins configurations

hakermi:
This PR should fix :
-Share document to user notifications : email, web and push
-Share document to space notifications : email web and push
-FileActivityChildPlugin configuration (still need updates).


ecm-wcm-extension/src/main/webapp/WEB-INF/conf/social/integration-configuration.xml
```diff
@@ -399,73 +411,6 @@
       </init-params>
     </component-plugin>
   </external-component-plugins>
-  <external-component-plugins>
```
ahamdi, line 402: @hakermi Just a question, you removed those plugins because they are duplicated ?

hakermi: yes, and that may caused a problem, thus the plugins can't locate the groovy templates

ahamdi (approved):

PR #1073, opened 2021-01-11
Task-39254: Display detected positive items on the quarantine page

AzmiTouil:


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -100,11 +100,18 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
     DlpPositiveItemService dlpPositiveItemService = CommonsUtils.getService(DlpPositiveItemService.class);
     DlpPositiveItemEntity dlpPositiveItemEntity = new DlpPositiveItemEntity();
     dlpPositiveItemEntity.setReference(node.getUUID());
+    dlpPositiveItemEntity.setTitle(node.getName());
+    if (node.hasProperty("exo:owner")) {
+      String author = node.getProperty("exo:owner").getString();
+      if ("__system".equals(author)) {
+        return;
```
rdenarie: Why do you return is the author is system ?
should we just continue to execute what we have after the if ?

apps/portlet-dlp-quarantine/src/main/webapp/vue-apps/dlpQuarantine/components/DlpQuarantineApp.vue
```diff
@@ -55,56 +66,62 @@
 </template>
 
 <script>
+import * as dlpQuarantineService from '../DlpQuarantineService';
```
rdenarie:
```suggestion
import * as dlpQuarantineService from '../dlpQuarantineService';
```

rdenarie (changes requested):
In addition of all comments, this PR should move in social

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
AzmiTouil: Currently, we add all the file in the dlp quarantine , I added this case because we have system files that are added too when the server starts up, I think they shouldn't.

rdenarie: In fact, theses items will not match the keywords so, they will not be put in DLPPositive
So, I think we should not return here if author is system

rdenarie (dismissed):

rdenarie (approved):

PR #1074, opened 2021-01-11
Task-38262 : Enable/Disable DLP feature

ahmadmido:


rdenarie (changes requested):
This PR sould move to social. 
Be careful, I think that there will be conflict with @AzmiTouil work

PR #1076, opened 2021-01-12
Move Dlp quarantine portlet to social

AzmiTouil:


rdenarie (approved):

rdenarie (changes requested):
Remove i18n from translation properties file

rdenarie (approved):

PR #1084, opened 2021-01-21
Task-37846 : Active DLP restore action

ahmadmido:


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -83,6 +92,27 @@ public boolean processItem(String entityId) {
     return true;
   }
 
+  @Override
+  public void movePositiveItem(String itemReference) {
```
rdenarie:
```suggestion
  public void restorePositiveItem(String itemReference) {
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -83,6 +92,27 @@ public boolean processItem(String entityId) {
...
+      long startTime = System.currentTimeMillis();
+      session = (ExtendedSession) WCMCoreUtils.getSystemSessionProvider().getSession(COLLABORATION_WS, repositoryService.getCurrentRepository());
+      Node node = session.getNodeByIdentifier(itemReference);
+      String fileName = node.getName();
+      restoreFromSecurity("/" + DLP_SECURITY_FOLDER+ "/" + fileName, WCMCoreUtils.getUserSessionProvider());
+      long endTime = System.currentTimeMillis();
+      long totalTime = endTime - startTime;
+      LOGGER.info("service={} operation={} parameters=\"fileName:{}\" status=ok " + "duration_ms={}",
+                  DlpOperationProcessor.DLP_FEATURE,
+                  "moveDlpPositiveDetection",
```
rdenarie:
```suggestion
                  "restoreItemDocument",
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -152,4 +183,108 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
...
+
+    try {
+      Integer.parseInt(path.substring(leftBracket+1, rightBracket));
+    } catch (Exception ex) {
+      return path;
+    }
+    return path.substring(0, leftBracket);
+  }
+
+  private void restoreFromSecurity(String SecurityNodePath,
```
rdenarie:
```suggestion
  private void restoreFromQuarantine(String SecurityNodePath,
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -152,4 +183,108 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
     dlpPositiveItemEntity.setKeywords(dlpKeywords);
     dlpPositiveItemService.addDlpPositiveItem(dlpPositiveItemEntity);
   }
+
+  private void removeMixinEXO_RESTORE_LOCATION(Session session, String restorePath) throws Exception {
```
rdenarie:
```suggestion
  private void removeRestorePathInfo(Session session, String restorePath) throws Exception {
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -152,4 +183,108 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
...
+    String restorePath = securityNode.getProperty(RESTORE_PATH).getString();
+
+    ManageableRepository manageableRepository = repositoryService.getCurrentRepository();
+    Session restoreSession = sessionProvider.getSession(restoreWorkspace,  manageableRepository);
+
+    if (restoreWorkspace.equals(securityWorkspace)) {
+      securityNodeSession.getWorkspace().move(SecurityNodePath, restorePath);
+    } else {
+      //clone node
+      restoreSession.getWorkspace().clone(
```
rdenarie: Does this case exists ?
I think the FileDlpAction should be configured only for collaboration workspace, so we cannot have a file which comes from another workspace

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/MoveFileDLPListener.java
```diff
@@ -0,0 +1,23 @@
+package org.exoplatform.ecm.connector.dlp;
+
+import org.exoplatform.commons.dlp.connector.DlpServiceConnector;
+import org.exoplatform.commons.dlp.processor.DlpOperationProcessor;
+import org.exoplatform.commons.dlp.service.DlpPositiveItemService;
+import org.exoplatform.commons.utils.CommonsUtils;
+import org.exoplatform.services.listener.Event;
+import org.exoplatform.services.listener.Listener;
+
+public class MoveFileDLPListener extends Listener<DlpPositiveItemService, String> {
```
rdenarie:
```suggestion
public class RestoreFileDLPListener extends Listener<DlpPositiveItemService, String> {
```

rdenarie (changes requested):
How to you ensure that, when restoring a document, it will not relaunch the action FileDLPAction ? and then be viewed as positive on more time, and immediatly moved in quarantine ?
Did you check this point ?
Does the FileDlpAction is relaunched after restoration ?

In addition, did you check that, after restore the file is correctly reindexed ?

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -13,20 +11,19 @@
 import org.exoplatform.commons.search.index.IndexingService;
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.commons.dlp.queue.QueueDlpService;
-import org.exoplatform.commons.search.index.IndexingService;
 import org.exoplatform.container.xml.InitParams;
 import org.exoplatform.container.xml.ValueParam;
+import org.exoplatform.services.cms.link.LinkManager;
```
rdenarie:
```suggestion
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -49,6 +46,16 @@
 
   private static final String OWNER                = "exo:owner";
 
+  final static public String EXO_RESTORE_LOCATION = "exo:restoreLocation";
+
+  final static public String RESTORE_PATH = "exo:restorePath";
+
+  final static public String RESTORE_WORKSPACE = "exo:restoreWorkspace";
+
+  final static public String EXO_TOTAL = "exo:total";
```
rdenarie:
```suggestion
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -152,4 +181,84 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
...
+                                    .getSession(COLLABORATION_WS,
+                                                repositoryService.getCurrentRepository());
+      return (Node) session.getItem("/" + DLP_SECURITY_FOLDER);
+    } catch (Exception e) {
+      return null;
+    }
+
+  }
+
+  private String fixRestorePath(String path) {
```
rdenarie, line 309: Can you explain what is this for ?

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -152,4 +181,84 @@ private void saveDlpPositiveItem(Node node) throws RepositoryException {
...
+
+    ManageableRepository manageableRepository = repositoryService.getCurrentRepository();
+    Session restoreSession = sessionProvider.getSession(restoreWorkspace,  manageableRepository);
+
+    if (restoreWorkspace.equals(securityWorkspace)) {
+      securityNodeSession.getWorkspace().move(SecurityNodePath, restorePath);
+    }
+    removeRestorePathInfo(restoreSession, restorePath);
+    securityHomeNode.save();
+    restoreSession.save();
```
rdenarie:
```suggestion
    if (restoreWorkspace.equals(securityWorkspace)) {
      securityNodeSession.getWorkspace().move(SecurityNodePath, restorePath);
      removeRestorePathInfo(restoreSession, restorePath);
      securityHomeNode.save();
      restoreSession.save();
    }
    
```

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
ahmadmido: This method is used to fix the path if it contains '[ 'or ']'.

rdenarie: In which case the path contains [ ?
And why should we fix that ?
What is not working ?

rdenarie (dismissed):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -105,11 +108,13 @@ public void restorePositiveItem(String itemReference) {
       Node node = session.getNodeByIdentifier(itemReference);
       String fileName = node.getName();
       restoreFromQuarantine("/" + DLP_SECURITY_FOLDER+ "/" + fileName, WCMCoreUtils.getUserSessionProvider());
+      indexingService.reindex(TYPE, itemReference);
```
rdenarie, line 107: I'm not sure we have to explicitly reindex it here. 
The reindexation should rerun automatically when the item is moved to original folder. My question is : we should ensure that it is really the case.

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -154,7 +159,8 @@ protected void treatItem(String entityId, Collection<SearchResult> searchResults
       Node node = session.getNodeByIdentifier(entityId);
       String fileName = node.getName();
       String restorePath = fixRestorePath(node.getPath());
-      if (!node.getPath().startsWith("/"+DLP_SECURITY_FOLDER+"/")) {
+      RestoredDlpItem restoredDlpItem = findRestoredDlpItem(node.getUUID());
+      if (!node.getPath().startsWith("/"+DLP_SECURITY_FOLDER+"/") && restoredDlpItem == null) {
```
rdenarie: If fact, we should not check only null for restoredDlpItem.

If the object restoreDLPItem is null => treat it, no problem
If the object restoreDLPItem is not null : 
if the node.getModifiedDate is after restoreDlpItem.getDate : it means that the document was modified AFTER his restoration. So that it could be different from the one restored, so that, we must treat it.

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -32,6 +32,8 @@
 import org.exoplatform.web.controller.metadata.ControllerDescriptor;
 import org.exoplatform.web.controller.router.Router;
 
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
```
rdenarie: Unecessary dependencies

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
rdenarie: cc @ahmadmido

rdenarie (changes requested):

rdenarie (dismissed):

rdenarie (approved):

PR #1086, opened 2021-01-22
Task-39798 : Add the mention external to the author of documents in the quarantine page

ahmadmido:


rdenarie (changes requested):
This PR is not needed. All stuff should be done at display

ahmadmido:
> This PR is not needed. All stuff should be done at display

This PR used to implement check external method, we can't check external in commons

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -176,6 +179,13 @@ public void removePositiveItem(String itemReference) {
     }
   }
 
+  @Override
+  public boolean checkExternal(String userId) {
+    IdentityManager identityManager = CommonsUtils.getService(IdentityManager.class);
+    Identity identity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, userId);
+    return identity.getProfile().getProperty("external") != null &&  identity.getProfile().getProperty("external").equals("true");
+  }
```
rdenarie, line 187: Should we put that in the parent class DlpConnector ?

rdenarie (changes requested):

rdenarie (approved):

PR #1089, opened 2021-01-25
Task-37850: Integrate the DLP keywords file via UI

AzmiTouil:


rdenarie (approved):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -98,6 +102,11 @@ private boolean isIndexedByEs(String entityId) {
 
   private void checkMatchKeywordAndTreatItem(String entityId) {
     SearchContext searchContext = null;
+    SettingService settingService = CommonsUtils.getService(SettingService.class);
+    SettingValue<?> settingValue = settingService.get(Context.GLOBAL, Scope.GLOBAL.id("DlpKeywords"), "exo:dlpKeywords");
+    if (settingValue != null) {
+      dlpKeywords = settingValue.getValue().toString().replace(",", " ");
+    }
```
rdenarie: Could we put that at upper level, in DlpConnector, with a function named getKeywords ?
So that, it could be exploited by all connectors implementation ?

In addition, I think that the variable dlpKeyword should be declared at global level.

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
AzmiTouil: Done!

rdenarie: I want to say 
the variable dlpKeyword should NOT be declared at global level.
Sorry

So to complete : getKeyword should be in DlpService
And we should NOT have a variable named dlpKeyword in FileDlpConnector

AzmiTouil: Done!

rdenarie (approved):

PR #1091, opened 2021-01-26
TASK-39859 Return HTTP 404 on OpenInOffice endpoint when user doesn't have access to node

boubaker:
Returning an HTTP code different from 404 gives the information to the user that the node exists with the given name. Thus this can be considered as a vulnerability

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
+++ b/core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
@@ -17,13 +17,16 @@ import org.json.JSONObject;
 import org.picocontainer.Startable;
 
 import javax.annotation.security.RolesAllowed;
-import javax.jcr.Node;
-import javax.jcr.PathNotFoundException;
-import javax.jcr.RepositoryException;
-import javax.jcr.Session;
+import javax.jcr.*;
+import javax.jcr.lock.LockException;
+import javax.jcr.nodetype.ConstraintViolationException;
+import javax.jcr.nodetype.NoSuchNodeTypeException;
+import javax.jcr.version.VersionException;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.*;
 import javax.ws.rs.core.*;
+import javax.ws.rs.core.Response.Status;
+
 import java.net.URLDecoder;
 import java.util.Locale;
 import java.util.MissingResourceException;
@@ -74,6 +77,7 @@ public class OpenInOfficeConnector implements ResourceContainer, Startable {
    * @throws Exception
    */
   @GET
+  @RolesAllowed("users")
   @Path("/updateDocumentTitle")
   public Response updateDocumentTitle(
           @Context Request request,
@@ -132,7 +136,9 @@ public class OpenInOfficeConnector implements ResourceContainer, Startable {
       if (linkManager.isLink(node)) node = linkManager.getTarget(node);
       nodePath = node.getPath();
       isFile = node.isNodeType(NodetypeConstant.NT_FILE);
-    }catch(RepositoryException ex){
+    } catch (PathNotFoundException | AccessDeniedException e) {
+      return Response.status(Status.NOT_FOUND).build();
+    } catch(RepositoryException ex){
       if(log.isErrorEnabled()){log.error("Exception when get node with path: "+filePath, ex);}
     }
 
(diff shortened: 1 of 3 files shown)


core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
```diff
@@ -132,7 +136,9 @@ public Response updateDocumentTitle(
       if (linkManager.isLink(node)) node = linkManager.getTarget(node);
       nodePath = node.getPath();
       isFile = node.isNodeType(NodetypeConstant.NT_FILE);
-    }catch(RepositoryException ex){
+    } catch (PathNotFoundException | AccessDeniedException e) {
+      return Response.status(Status.NOT_FOUND).build();
```
plamarque, line 146: Shouldn't we log something in this case ?

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
```diff
@@ -183,28 +189,40 @@ public Response updateDocumentTitle(
...
+      }
 
-    return Response.ok(String.valueOf(node.isCheckedOut()), MediaType.TEXT_PLAIN).build();
+      if (!node.isCheckedOut()) {
+        node.checkout();
+      }
+
+      return Response.ok(String.valueOf(node.isCheckedOut()), MediaType.TEXT_PLAIN).build();
+    } catch (PathNotFoundException | AccessDeniedException e) {
+      return Response.status(Status.NOT_FOUND).build();
```
plamarque, line 231: Shouldn't we log something in this case ?

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
```diff
@@ -216,26 +234,38 @@ public Response checkout(@Context Request request,
...
+    try {
+      Node node = null;
+      try {
+        node = (Node) nodeFinder.getItem(workspace, filePath);
+      } catch (PathNotFoundException e) {
+        node = (Node) nodeFinder.getItem(workspace, Text.unescapeIllegalJcrChars(filePath));
+      }
+      String repo = WCMCoreUtils.getRepository().getConfiguration().getName();
+
+      String obsPath = httpServletRequest.getScheme() + "://" + httpServletRequest.getServerName() + ":"
```
plamarque, line 263: Shouldn't this be /portal/rest instead of /rest ?

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
```diff
@@ -216,26 +234,38 @@ public Response checkout(@Context Request request,
...
+          + WCMCoreUtils.getRestContextName() + "/private/jcr/" + repo + "/" + workspace + node.getPath();
+
+      String shortCutContent = "[InternetShortcut]\n";
+      shortCutContent += "URL=" + obsPath + "\n";
+      return Response.ok(IOUtils.toInputStream(shortCutContent), MediaType.APPLICATION_OCTET_STREAM)
+                     .header("Content-Disposition", "attachment; filename=" + node.getName() + ".url")
+                     .header("Content-type", "application/internet-shortcut")
+                     .build();
+    } catch (PathNotFoundException | AccessDeniedException e) {
+      return Response.status(Status.NOT_FOUND).build();
```
plamarque, line 278: Shouldn't we log something in this case ?

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
boubaker: Nope, not needed, we will not have to log any request that returns a regular and normal 404 HTTP code. (We have frontend access log for that)

boubaker: ok, will add debug log for that in case Ops wants to see entries of 404 here instead of access log of frontend

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
boubaker: Nope, this is needed with Basic authentication

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
boubaker: done

core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/OpenInOfficeConnector.java
boubaker: done

ahamdi (approved):
LGTM
I reverted https://github.com/exoplatform/ecms/pull/1092 as it is included in this PR

PR #1101, opened 2021-02-01
Task-40167: Fix display keyword When the detected keyword on the document contains a capital letter

AzmiTouil:


rdenarie (changes requested):
Does your modification works with accentued char ?
Example : in document : "accentué", and in keyword : "accentue"

AzmiTouil:
> Does your modification works with accentued char ?
> Example : in document : "accentué", and in keyword : "accentue"

no , we need to remove accents

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -208,12 +210,20 @@ private String getDetectedKeywords(Collection<SearchResult> searchResults, Strin
...
                      }
                    }
                  });
     return detectedKeywords.stream().collect(Collectors.joining(", "));
   }
+
+  private String removeAccents(String string) {
+    if (StringUtils.isNotBlank(string)) {
+      string = Normalizer.normalize(string, Normalizer.Form.NFD);
+      string = string.replaceAll("[\\p{InCombiningDiacriticalMarks}]", "");
```
rdenarie, line 225: are you sure of this line ?

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
AzmiTouil: Yes

rdenarie: Can you explain where it comes from ?

AzmiTouil: The \p{InCombiningDiacriticalMarks} will match all such diacritic codes and we will replace them with their deAccented counterparts followed by their combining diacritics.
we have already used it in other places https://github.com/Meeds-io/social/blob/d72cad054a333238e99f0e638f271fd9cc94555c/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java#L443

rdenarie (approved):

PR #1110, opened 2021-02-04
Removed restrictive Google Drive scopes in Cloud Drive

pnedonosko:
Removed Google Drive scopes in Cloud Drive connector: `drive.scripts`, `drive.apps.readonly`.


pnedonosko:
Wrong commits in the PR

PR #1114, opened 2021-02-05
Task-40049: Implement user notifications when Dlp item are restored

AzmiTouil:


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/RestoreFileDLPListener.java
```diff
@@ -1,23 +1,23 @@
...
 import org.exoplatform.commons.dlp.connector.DlpServiceConnector;
+import org.exoplatform.commons.dlp.domain.DlpPositiveItemEntity;
 import org.exoplatform.commons.dlp.processor.DlpOperationProcessor;
 import org.exoplatform.commons.dlp.service.DlpPositiveItemService;
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.services.listener.Event;
 import org.exoplatform.services.listener.Listener;
 
-public class RestoreFileDLPListener extends Listener<DlpPositiveItemService, String> {
+public class RestoreFileDLPListener extends Listener<DlpPositiveItemService, Object> {
```
rdenarie:
```suggestion
public class RestoreFileDLPListener extends Listener<DlpPositiveItemService, DlpPositiveItemEntity> {
```
I think it is better to use the object directly instead of casting it ?
WDYT ?

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/RestoreFileDLPListener.java
AzmiTouil: Yes , better !

rdenarie (approved):

PR #1115, opened 2021-02-05
Task-40255 : [DLP] a warning is displayed when the admin tries to view a document from the quarantine page

rdenarie:
Before this fix, a popup is displayed when an admin tries to open a document in the quarantine folder
This problem is due to the fact that UIJCRExplorerPortlet checks node before displaying, and if the node have restoreLocation properties, it means that it is in trash
In our case, the node is not in trash and should be displayed.

This fix add a check to see if the node is in trash or not. If yes, nothing change.
If not, we check if we are speaking about the real node a linked node. If it is the real, we display it (admin read it in quarantine folder).
If it is a link, we do not allow. It is a case a user read a link to realNode which is quarantine, so we should not display it.


azayati (approved):
LGTM

PR #1116, opened 2021-02-05
Task-40386 : Bug : [DLP] Restored documents are not available for the author

rdenarie:
Beofre this fix, when a file is restored, restore properties and mixin are not removed
This ensure that the mixin is correctly removed


azayati (approved):
LGTM

PR #1123, opened 2021-02-15
Task-40572: avoid dlp item processing when updating it via co-editing

AzmiTouil:
Bugs:
**1-** Once a document is restored, when the user updates it via the coediting option and adds new keywords, the document is sent to quarantine but the detected keywords are not displayed ==> We should display the newly detected keywords

**2-** Once a document is restored, when the user opens the document in coedition, this document is sent to the quarantine page even if the user doesn't change the document content ==> The document must be sent to the quarantine page only if the new version contains keywords

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
+++ b/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
@@ -31,9 +31,7 @@ public class FileDLPAction implements AdvancedAction {
   private ExoFeatureService featureService;
   
   private static final String EXO_EDITORS_RUNTIME_ID = "exo:editorsId";
-
-  private static final String EXO_CURRENT_PROVIDER   = "exo:currentProvider";
-
+ 
   public FileDLPAction() {
     this.trashService = CommonsUtils.getService(TrashService.class);
     this.queueDlpService = CommonsUtils.getService(QueueDlpService.class);
@@ -61,7 +59,7 @@ public class FileDLPAction implements AdvancedAction {
       case Event.PROPERTY_ADDED:
       case Event.PROPERTY_CHANGED:
         PropertyImpl property = (PropertyImpl) context.get(InvocationContext.CURRENT_ITEM);
-        if (property != null && !property.getName().equals(EXO_EDITORS_RUNTIME_ID) && !property.getName().equals(EXO_CURRENT_PROVIDER) && !property.getName().equals(FileDlpConnector.RESTORE_WORKSPACE) && !property.getName().equals(FileDlpConnector.RESTORE_PATH)) {
+        if (property != null && !property.getName().equals(EXO_EDITORS_RUNTIME_ID) && !property.getName().equals(FileDlpConnector.EXO_CURRENT_PROVIDER) && !property.getName().equals(FileDlpConnector.RESTORE_WORKSPACE) && !property.getName().equals(FileDlpConnector.RESTORE_PATH)) {
           node = property.getParent();
           if (node != null && !trashService.isInTrash(node)) {
             if (node.isNodeType(NodetypeConstant.NT_RESOURCE)) {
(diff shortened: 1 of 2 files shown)


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
```diff
@@ -61,7 +59,7 @@ public boolean execute(Context context) throws Exception {
       case Event.PROPERTY_ADDED:
       case Event.PROPERTY_CHANGED:
         PropertyImpl property = (PropertyImpl) context.get(InvocationContext.CURRENT_ITEM);
-        if(property != null && !property.getName().equals(EXO_EDITORS_RUNTIME_ID) && !property.getName().equals(FileDlpConnector.RESTORE_WORKSPACE) && !property.getName().equals(FileDlpConnector.RESTORE_PATH)) {
+        if (property != null && !property.getName().equals(EXO_EDITORS_RUNTIME_ID) && !property.getName().equals(FileDlpConnector.EXO_CURRENT_PROVIDER) && !property.getName().equals(FileDlpConnector.RESTORE_WORKSPACE) && !property.getName().equals(FileDlpConnector.RESTORE_PATH)) {
```
rdenarie: This modification should not be necessary

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
AzmiTouil: It is necessary, because if we delete that, the file will be quarantined once we open and close the editor without making any changes to the file.

rdenarie: Ok. Can you rework this part and create a list of excluded property names ? 
So that the if will be 
if (property!=null && !excludedPropertyNames.contains(property.getName()) {
...

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
```diff
@@ -31,6 +35,11 @@
   private ExoFeatureService featureService;
   
   private static final String EXO_EDITORS_RUNTIME_ID = "exo:editorsId";
+
+  private static final List<String> EXCLUDED_PROPERTY_NAMES   = Collections.unmodifiableList(Arrays.asList(EXO_EDITORS_RUNTIME_ID,
```
rdenarie:
```suggestion
  private static final List<String> excludedPropertyNames   = Collections.unmodifiableList(Arrays.asList(EXO_EDITORS_RUNTIME_ID,
```

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDLPAction.java
```diff
@@ -59,7 +68,7 @@ public boolean execute(Context context) throws Exception {
       case Event.PROPERTY_ADDED:
       case Event.PROPERTY_CHANGED:
         PropertyImpl property = (PropertyImpl) context.get(InvocationContext.CURRENT_ITEM);
-        if (property != null && !property.getName().equals(EXO_EDITORS_RUNTIME_ID) && !property.getName().equals(FileDlpConnector.EXO_CURRENT_PROVIDER) && !property.getName().equals(FileDlpConnector.RESTORE_WORKSPACE) && !property.getName().equals(FileDlpConnector.RESTORE_PATH)) {
+        if (property != null && !EXCLUDED_PROPERTY_NAMES.contains(property.getName())) {
```
rdenarie:
```suggestion
        if (property != null && !excludedPropertyNames.contains(property.getName())) {
```

rdenarie (changes requested):

rdenarie (approved):

PR #1138, opened 2021-02-17
TASK-36121: Fix the document preview display

hassenelkamel-exo:
**Problem:** In order to fix display documents with onlyoffice in stream  we [PR](https://github.com/exoplatform/onlyoffice/pull/69) have removed height from only office,the problem in stream was fixed but it results a regression in the preview,
**Solution:** so to solve this problem we add height UIDocumentPreview to  insure that we display correctly in preview.
**IT will fix:**  the display in stream, preview embedded in documents app


plamarque (commented):
not sure if it is in the scope of this fix, but we also have a heigh problem in preview embedded in Documents app.

hassenelkamel-exo:
it will be fixed by this PR also @plamarque

boubaker (approved):

PR #1141, opened 2021-02-18
Task-40529: Avoid displaying quarantined item in AS

AzmiTouil:
Prior to this fix, when uploading a file attached to an activity, and it contains dlp keywords, it will be quarantined but it is still displayed in AS.

This fix will ensure that an quarantined file is not displayed until validated.


ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/FileUIActivity.java
```diff
@@ -883,6 +883,10 @@ public void setUIActivityData(Map<String, String> activityParams) {
     return (valuesArray == null || index > (valuesArray.length - 1)) ? null : valuesArray[index];
   }
 
+  private boolean isQuarantinedItem(Node node) throws RepositoryException {
+    return node.getPath().startsWith("/Security");
```
rdenarie: Can you use FileDLPConnector.DLP_SECURITY_FOLDER ? 
As we will change this name in future, it iwll prevent us to forget this one :)

rdenarie (changes requested):

ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/FileUIActivity.java
AzmiTouil: Done!

rdenarie (approved):

PR #1144, opened 2021-02-19
Maint-39705 : Upload new version is not working when the file contains a space in i…

Rdidmohamedelhadi:
…ts name

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/RenameConnector.java
+++ b/core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/RenameConnector.java
@@ -45,6 +45,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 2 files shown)


core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/RenameConnector.java
```diff
@@ -110,7 +111,9 @@ public Response rename(@QueryParam("oldPath") String oldPath,
       }
       String newExoTitle = newTitle;
       // Clarify new name & check to add extension
-      String newName = Text.escapeIllegalJcrChars(Utils.cleanString(newTitle));
+      String newName  = Text.escapeIllegalJcrChars(newTitle);
```
boubaker, line 114: `Text.escapeIllegalJcrChars` should be sufficient to remove accent and other characters.

Rdidmohamedelhadi: ok

ahamdi (approved):

PR #1157, opened 2021-02-22
Task-40565 :  Favorites documents in quarantine should not be accessible

ahmadmido:
- Before this fix favorite documents are displayed in the favorite widget on the snapshot page when are detected by quarantine and the symlinks also are displayed on the favorite folder without permissions to access.
- This fix will assure that documents on the quarantine page aren't displayed on the snapshot page and the symlinks also are deleted.

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
+++ b/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
@@ -19,6 +19,7 @@ import org.exoplatform.commons.dlp.queue.QueueDlpService;
 import org.exoplatform.container.xml.InitParams;
 import org.exoplatform.services.cms.drives.DriveData;
 import org.exoplatform.services.cms.drives.ManageDriveService;
+import org.exoplatform.services.cms.impl.Utils;
 import org.exoplatform.services.jcr.RepositoryService;
 import org.exoplatform.services.jcr.access.PermissionType;
 import org.exoplatform.services.jcr.core.ExtendedNode;
@@ -163,6 +164,7 @@ public class FileDlpConnector extends DlpServiceConnector {
       if (!node.getPath().startsWith("/"+DLP_SECURITY_FOLDER+"/") && (restoredDlpItem == null ||  getNodeLastModifiedDate(node) > restoredDlpItem.getDetectionDate())) {
         workspace.move(node.getPath(), "/" + DLP_SECURITY_FOLDER + "/" + fileName);
         indexingService.unindex(TYPE, entityId);
+        Utils.removeDeadSymlinks(node);
         saveDlpPositiveItem(node,searchResults);
         addRestorePathInfo(node.getName(), restorePath, workspace.getName());
         long endTime = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -167,6 +168,7 @@ protected void treatItem(String entityId, Collection<SearchResult> searchResults
       RestoredDlpItem restoredDlpItem = findRestoredDlpItem(node.getUUID());
       if (!node.getPath().startsWith("/"+DLP_SECURITY_FOLDER+"/") && (restoredDlpItem == null ||  getNodeLastModifiedDate(node) > restoredDlpItem.getDetectionDate())) {
         workspace.move(node.getPath(), "/" + DLP_SECURITY_FOLDER + "/" + fileName);
+        Utils.removeDeadSymlinks(node);
```
rdenarie: I'm not sure that we should remove symlink. 
When we keep the symlink, and when the target document goes in quarantine, what is the behaviour when you click on the symlink ?

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
```diff
@@ -899,7 +900,9 @@ public long getEditorsIdleTimeout() {
       if (userPrivateNode.hasNode(NodetypeConstant.FAVORITE)) {
         favoriteFolder = ((Node) userPrivateNode.getNode(NodetypeConstant.FAVORITE)).getPath();
       }
-      return getDocumentsByFolder(favoriteFolder, null, limit);
+      List<Document> favoriteDocuments = getDocumentsByFolder(favoriteFolder, null, limit);
```
rdenarie: Why have we to filter on document in Security folder ?
Why documents in trash are not displayed ? How ? There is no filter to know if a document is in trash

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
ahmadmido: I have discussed with Sameh and this the normal behavior is to remove symlinks when the document is moved to the quarantine. The symlink is not accessible when the document is moved to the quarantine page.

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
ahmadmido: This test is used to check documents on the snapshot page favorite documents.

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
rdenarie: Why ? When the document is restored, the symlinks are not restored. 
Can you try to answer to this  :
When we dont remove symlink,  what is the behaviour when you click on a symlink of a document which is in quarantine ?

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
rdenarie: You do not answer to my question :)
Why documents in trash are not displayed ? How ? There is no filter to know if a document is in trash
Why there is not filter to trash ? 

I want to understand how it work for trash to be sure to be consistent.

My idea is that for trash, favorites are removed, so, there is not need to add a filter. Can you confirm ?

rdenarie (dismissed):

rdenarie (approved):

PR #1166, opened 2021-02-25
Task-41054: Fix keywords edition feedbacks

AzmiTouil:


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -294,13 +294,13 @@ private String getDetectedKeywords(Collection<SearchResult> searchResults, Strin
...
   private String escapeSpecialCharacters(String keyword) {
-    List<String> keywordParts = Arrays.stream(keyword.split("[+\\-=&|><!(){}\\[\\]^\"*?:/ @$]+"))
+    List<String> keywordParts = Arrays.stream(keyword.split("[+\\-=&|><!(){}\\[\\]^\"*?:/ @$#]+"))
                                       .distinct()
                                       .collect(Collectors.toList());
     for (String s : keywordParts) {
       keyword = keyword.replace(s, "<em>" + s + "</em>");
     }
-    return keyword;
+    return keyword.replaceAll("'","’");
```
rdenarie, line 271: Can you explain why you replace ' by ’ ? why it is necessary ? I afraid that it will add a non standard character

rdenarie (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
AzmiTouil: I replaced it, because ES indexes all files with ’ instead of ' , so I need to replace it too , to find the detected keywords

rdenarie: Can you add a comment then, to explain why this replace is needed ?

rdenarie (approved):

PR #1168, opened 2021-02-26
TASK-40565 : Favorite documents in quarantine shouldn't be accessible

ahmadmido:
- Before this fix favorite items are displayed in the favorite widget on the snapshot page when are detected by quarantine and when we delete the item from the quarantine page the dead symlinks are displayed.
- This fix will assure that items on the quarantine page aren't displayed on the snapshot page and dead symlinks also are deleted by adding a mixin when the item is moved to the quarantine page to all symlinks and we remove mixin after restoring action.
- Dead symlinks also are deleted when the item is deleted on the quarantine page.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
@@ -828,6 +828,7 @@ public class DocumentServiceImpl implements DocumentService {
          + "or exo:fileType like '%odt%' "
          + "or exo:fileType like '%officedocument%' "
          + "or exo:fileType like '%ms%') "
+         + "and not (jcr:mixinTypes like '%exo:exoDlpItem%' )"
          + (condition != null ? condition : "")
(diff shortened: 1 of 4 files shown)


core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
```diff
@@ -828,6 +828,7 @@ public long getEditorsIdleTimeout() {
          + "or exo:fileType like '%odt%' "
          + "or exo:fileType like '%officedocument%' "
          + "or exo:fileType like '%ms%') "
+         + "and not (exo:dlp = 'dlpsymlink' )"
```
rdenarie:
```suggestion
         + "and not (jcr:primaryType = 'exo:exoDlpItem' )"
```
Thus I'm not sure that we need a specific property. The mixin should be enough

rdenarie (changes requested):

rdenarie (dismissed):

rdenarie (dismissed):

rdenarie (approved):

PR #1175, opened 2021-03-03
MAINT-41089: Make sure that the title of uploaded file with special characters is not encoded.

mdiha:
When uploading a file that it's title contains special characters it is changed by replacing the special characters by the character (-) due to JCR limitation, these changes are required for the created node name property but it is not required for the title so i made sure to apply these changes only for the file's node name by keeping its title as the original value.


ahamdi (approved):

ahamdi (changes requested):
Description required

mdiha:
> Description required

Description added

ahamdi:
Did you test for cyrillic characters ?

mdiha:
> Did you test for cyrillic characters ?

I did test it for cyrillic characters and it works fine

ahamdi (approved):

PR #1176, opened 2021-03-04
Task-41197: Remove log warn related to doc breadcrumb when user is anonymous

AzmiTouil:
Before this fix, frequently we have this warn in logs : 
"Can't find associated drive of node with path"  related to dlp process

This fix will remove log warn related to doc breadcrum when user is anonymous


rdenarie (dismissed):

rdenarie (approved):

azayati (approved):
LGTM

PR #1182, opened 2021-03-09
TASK-41334: fixed new drives are not displayed when we add a user in a new group

GouadriaHanen:
- on each add/remove of a user in a group the listener clears the cache of groups except generalDrives (ALL_MAIN_CACHED_DRIVE )
- clear cache of generalDrives (ALL_MAIN_CACHED_DRIVE ) on add/remove a user in a group


ahamdi (approved):
LGTM
Good job !

PR #1188, opened 2021-03-11
FileDlpConnector : improve error log

rdenarie:
Before this commit, when the function throws an error, we dont know which item is impacted
This fix improve the log to display it


azayati (approved):
LGTM

PR #1197, opened 2021-03-16
Task-40615: Fix activity generated and displayed on the Activity Stream once creating or updating a web-content by a web-contributor

NaderSai:
Problem: the activity stream was generated once creating a web content, because the property of generating activity is enabled.
How it was solved : disable the property of generating activity from entreprise-configuration.properties.


ahamdi:
This is not what is requested by the task, please check @NaderSai

ahamdi (approved):

PR #1198, opened 2021-03-17
fix the display of replace option instead of upload new version for documents with % char in name

hakermi:
This PR should fix the display of replace option instead of upload new version for documents with % char in name in the upload suggestion


ahamdi (approved):

hbenali (commented):
Please indicate the task in the PR's title

PR #1206, opened 2021-03-22
MAINT-41748/41778 : fix uploading documents wih % char is their names

hakermi:
fix versionning for document with % chat and their names

fix upload new version of document with % char in its name

fix the display of replace option instead of upload new version for documents with % char in name


ahamdi (approved):
LGTM

PR #1208, opened 2021-03-22
MAINT-41089: Make sure that the title of uploaded file with special c…

hakermi:
…haracters is not encoded. (#1175) (#1184)

When uploading a file that it's title contains special characters it is changed by replacing the special characters by the character (-) due to JCR limitation, these changes are required for the created node name property but it is not required for the title so i made sure to apply these changes only for the file's node name by keeping its title as the original value.


ahamdi (approved):
LGTM

PR #1219, opened 2021-03-31
Ensure that no element are added to queue when no keywords configured

rdenarie:
Before this fix, when no keywords are configured, item are queued, and fill the queue for nothing
This fix add a check to verify if there is at least one keyword.


azayati (approved):
LGTM

PR #1220, opened 2021-04-01
Task-42165: Fix Access to folders to which I have no permission to access

NaderSai:
Problem:  the current behavior of UIOnePathSelector was selecting from all reprositories.
How it was solved: by changing this behavior to select from drives instead of selecting a node/content from all reprositories.


ahamdi (changes requested):
There are many unneded changes in the PR, could you fix them?
 and update the description with the problem and the provided solution. You changed the current behavior of UIOnePathSelector component to select from drives instead of selecting a node/content from all reprositories.

ahamdi (approved):
LGTM

PR #1229, opened 2021-04-15
TASK-41656: add new Attachments App in snapshot page

aelbenney:


boubaker (approved):

boubaker:
We will for sure refactor the code to make small changes Vue components instead of huge one. In addition, we will have to take care of RTL feature from now on ;-)
But let's make a first version that works fine in functional perspective, we will enhance this in a second phase.

PR #1234, opened 2021-04-21
Task-42330 : [DLP] Some items are stuck in the queue

rdenarie:
Before this commit, items which no more exists (not only in trash but completly removed) throw an error, so that they are not removed from queue
This commit add a test to check if an item exists, if not, it is removed from the queue.
In addition this commit add a debug to display the path of an entityId (we need that for other items stucks in the queue)

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
+++ b/core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
@@ -105,7 +105,7 @@ public class FileDlpConnector extends DlpServiceConnector {
   }
   
   private boolean itemExist(String entityId) {
-    ExtendedSession session;
+    ExtendedSession session = null;
     boolean result = false;
     try {
       session = (ExtendedSession) WCMCoreUtils.getSystemSessionProvider().getSession(COLLABORATION_WS, repositoryService.getCurrentRepository());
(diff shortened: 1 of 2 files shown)


core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
```diff
@@ -104,6 +104,23 @@ public boolean processItem(String entityId) {
     return true;
   }
   
+  private boolean itemExist(String entityId) {
+    ExtendedSession session;
+    boolean result = false;
+    try {
+      session = (ExtendedSession) WCMCoreUtils.getSystemSessionProvider().getSession(COLLABORATION_WS, repositoryService.getCurrentRepository());
```
AzmiTouil, line 111: Please close the session at the end of this method

AzmiTouil (changes requested):

core/connector/src/main/java/org/exoplatform/ecm/connector/dlp/FileDlpConnector.java
rdenarie: As we use the sessionProvider which use a cache for session, we dont need to close the session.
The session will be used (and will be the same as long as it is valid). When this session will timeout, it will be closed and removed from the cache.
The sessionprovider manages this session.

AzmiTouil: Yes I know . But we have encountered some problems before 
https://github.com/exoplatform/ecms/commit/f58ac94ae7837dec6c20299d939b654338efa7e0

rdenarie: ok, updated

AzmiTouil (approved):
LGTM

ahmadmido (approved):
LGTM

PR #1236, opened 2021-04-26
Task-43025 : Documentary folders aren't accessible due to pagination settings

Rdidmohamedelhadi:
when we use the system provider, we will select all folders without taking into consideration the permission of the current user, so when we find that some folders are not accessible by the current user, we will not display all folders, so my fix is simply to select only the current user folders

### DIFF
--- a/core/search/src/main/java/org/exoplatform/services/wcm/search/base/LazyPageList.java
+++ b/core/search/src/main/java/org/exoplatform/services/wcm/search/base/LazyPageList.java
@@ -16,22 +16,20 @@
  */
 package org.exoplatform.services.wcm.search.base;
 
-import java.util.ArrayList;
-import java.util.List;
+import org.exoplatform.commons.utils.PageList;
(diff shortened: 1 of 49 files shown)


core/search/src/main/java/org/exoplatform/services/wcm/search/base/LazyPageList.java
```diff
@@ -87,7 +85,7 @@ protected void populateCurrentPage(int page) throws Exception {
     currentListPage_ = new ArrayList<E>();
     if (queryData_.getQueryStatement() != null) {
       try {
-        SessionProvider provider = WCMCoreUtils.getSystemSessionProvider();
+        SessionProvider provider = WCMCoreUtils.getUserSessionProvider();
```
ahamdi: Did you check if this class is used just for this case? olr is it used by other calsses for other documents paginating cases ? This change may cause regressions .

ahamdi (approved):
LGTM
We need to make sure that there are no regressions caused by this change. @eerrokh @mzorai

PR #1244, opened 2021-05-06
TASK-43877: Make attachments app reusable

aelbenney:
This PR makes the attachments drawer reusable as you can see your attachments if exist at first. Then attach new ones in the default path you have already configured. In order to use the attachments app in other context, you only need to add the vue component in your application and configure the  properties needed `(entityId, entityType, defaultDrive, defaultFolder, spaceId)`.


core/services/src/main/java/org/exoplatform/services/attachments/model/Attachment.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.LinkedHashMap;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class Attachment implements Cloneable {
+  private String id;
```
boubaker: code format

core/services/src/main/java/org/exoplatform/services/attachments/rest/model/AttachmentEntity.java
```diff
@@ -0,0 +1,50 @@
...
+import org.exoplatform.services.attachments.model.Permission;
+import org.exoplatform.social.rest.entity.IdentityEntity;
+
+import java.util.LinkedHashMap;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class AttachmentEntity {
+  private String id;
```
boubaker: code format

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -0,0 +1,306 @@
+/**
+ *
+ */
```
boubaker, line 16: empty comment, to replace with Licence Header for all new files

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Link an existing attachments to the given entity (Event, Task, Wiki,...)
+   *
+   * @param entityId the entity id.
+   * @param entityType the entity type.
+   * @param attachmentIds the list of attachments uuid to be linked.
+   * @return The response.
+   * @throws Exception The exception
+   * @anchor ManageDocumentService.linkAttachmentsToEntity
```
boubaker: No need to add javadoc API in REST endpoints since you have added swager comments

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentsService.java
```diff
@@ -0,0 +1,19 @@
+package org.exoplatform.services.attachments.service;
+
+import org.exoplatform.commons.exception.ObjectNotFoundException;
+import org.exoplatform.services.attachments.model.Attachment;
+
+import java.util.List;
+
+public interface AttachmentsService {
```
boubaker: Rename Class to use singular form please

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentsServiceImpl.java
```diff
@@ -0,0 +1,164 @@
...
+import org.exoplatform.services.jcr.RepositoryService;
+import org.exoplatform.services.jcr.ext.app.SessionProviderService;
+import org.exoplatform.services.jcr.ext.common.SessionProvider;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
+import javax.jcr.Session;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class AttachmentsServiceImpl implements AttachmentsService {
```
boubaker: Rename Class to use singular form

core/services/src/main/java/org/exoplatform/services/attachments/storage/AttachmentsStorage.java
```diff
@@ -0,0 +1,12 @@
+package org.exoplatform.services.attachments.storage;
+
+import org.exoplatform.services.attachments.model.AttachmentsContextEntity;
+
+public interface AttachmentsStorage {
```
boubaker: Rename Class to use singular form

core/services/pom.xml
```diff
@@ -11,6 +11,7 @@
   <description>eXo CMS Service</description>
   <properties>
     <exo.test.coverage.ratio>0.34</exo.test.coverage.ratio>
+    <org.lombok.version>1.18.2</org.lombok.version>
```
boubaker: please move version declaration to https://github.com/Meeds-io/maven-depmgt-pom
we will need to apply this change on other projects but not in the context of current task.

core/services/src/main/java/org/exoplatform/services/attachments/dao/AttachmentsDAO.java
```diff
@@ -0,0 +1,22 @@
+package org.exoplatform.services.attachments.dao;
+
+import org.exoplatform.commons.persistence.impl.GenericDAOJPAImpl;
+import org.exoplatform.services.attachments.model.AttachmentsContextEntity;
+
+import javax.persistence.NoResultException;
+import javax.persistence.TypedQuery;
+
+public class AttachmentsDAO extends GenericDAOJPAImpl<AttachmentsContextEntity, Long> {
```
boubaker: Use singular for class name

core/services/src/main/java/org/exoplatform/services/attachments/model/AttachmentsEntityType.java
```diff
@@ -0,0 +1,5 @@
+package org.exoplatform.services.attachments.model;
+
+public enum AttachmentsEntityType {
+  WIKI, EVENT, TASK;
```
boubaker: Attachment types should be extensible through a component plugin.
We should be able to attach files to any kind of applications without having to change code provided in ecms.

core/services/src/main/java/org/exoplatform/services/attachments/model/AttachmentsContextEntity.java
```diff
@@ -0,0 +1,69 @@
...
+  private Long id;
+
+  @Column(name = "ATTACHMENT_IDS")
+  private String attachmentIds;
+
+  @Column(name = "ENTITY_ID")
+  private Long entityId;
+
+  @Column(name = "ENTITY_TYPE")
+  private int entityType;
```
boubaker: The entity type can be of type `int` if you inject attachment types by component plugin and associate for each type of application a unique identifier that will be used here in this field. You can also, make this a String (simply), without having to add a dedicated component plugin for association. (I advice, for simplicity, to use a String)

core/services/pom.xml
aelbenney: okay

aelbenney: https://github.com/Meeds-io/maven-depmgt-pom/pull/34

core/services/src/main/java/org/exoplatform/services/attachments/model/AttachmentsEntityType.java
boubaker: example

> https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/spaces-navigation/initComponents.js#L16-L24
> https://github.com/Meeds-io/gatein-portal/blob/develop/portlet/web/src/main/webapp/vue-apps/hamburger-menu/components/ExoHamburgerMenuNavigation.vue#L129

core/services/src/main/java/org/exoplatform/services/attachments/model/AttachmentsContextEntity.java
aelbenney: done

boubaker (approved):

PR #1245, opened 2021-05-06
TASK-42100 : fixed erroned path redirection when opening two different documents in two different tabs

nidhalnidhal:
before this fix , when opening 2 documents from 2 different folders in 2 different tabs , the redirection from one document leads to the the other file parent folder instead of his own parent folder due to the fact that the object storing the current path depends on the session object so it returns the folder of the last file being opened . i fixed this by retrieving the previous path depending on the value of the current path Input from the UI.

### DIFF
--- a/core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/control/UIAddressBar.java
+++ b/core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/control/UIAddressBar.java
@@ -158,7 +158,10 @@ public class UIAddressBar extends UIForm {
           uiExplorer.setIsViewTag(false) ;
           uiExplorer.setSelectNode(uiExplorer.getCurrentStatePath()) ;
         } else {
-          String previousNodePath = uiExplorer.rewind() ;
+          // Retrieve previous node path by refactoring current path and adding the RootPath
+          String CurrentPath =  Text.escapeIllegalJcrChars(uiAddressBar.getUIStringInput(FIELD_ADDRESS).getValue());
+          String RootPath = uiExplorer.getRootPath();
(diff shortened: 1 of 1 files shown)


core/webui-explorer/src/main/java/org/exoplatform/ecm/webui/component/explorer/control/UIAddressBar.java
```diff
@@ -158,7 +158,10 @@ public void execute(Event<UIAddressBar> event) throws Exception {
           uiExplorer.setIsViewTag(false) ;
           uiExplorer.setSelectNode(uiExplorer.getCurrentStatePath()) ;
         } else {
-          String previousNodePath = uiExplorer.rewind() ;
+          // Retrieve previous node path by refactoring current path and adding the RootPath
+          String CurrentPath =  Text.escapeIllegalJcrChars(uiAddressBar.getUIStringInput(FIELD_ADDRESS).getValue());
+          String RootPath = uiExplorer.getRootPath();
```
ahamdi, line 163: You are using an object from the session here. Let's do a test, could you open two Documents applications in two pages for two different spaces , or one for Space Documents app and the other for the Personal documents app.
`String RootPath = uiExplorer.getRootPath();`

nidhalnidhal: @ahamdi , actually in case we open 2 documents from 2 different spaces in 2 different tabs , the problem is not reproduced . i tested it several times.

nidhalnidhal: and even when we open a document from personal documents app

ahamdi (approved):

PR #1248, opened 2021-05-20
MAINT-40548 : Fix the display of shared folder in the activity with attached files in a space

hakermi:
**ISSUE** : When create an activity with attached files in a space  the `DocumetService.getDocumentUrlInPersonalDocuments` is called inside the UI template and also in the plugin notification template which invoke  the sub `getSharedLink` method which verify the existing of the shared folder and and searching the file inside it and if it's not exist its create it inside the shared folder : `linkManager.createLink(shared, currentNode);`
This happens after storing the file inside the activity stream folder, and the link value is transmitted to the template.
And this is not the expected behavior, the shared folder should include only the shared files in the current space.
**Solution** : 
- Replace the create of the symlink inside the shared folder by searching it in Activity Stream folder and sends its link.
- **Fix a discovered issue** inside the `ActivityListener` which wasn't working properly (this has no relation with the initial issue)
      - The Activity files received when calling `activity.getFiles();` has a `null id` which throws an **NPE** while getting their nodes 
       ` session.getNodeByUUID(fileToShare.getId());`

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
@@ -138,6 +138,7 @@ public class DocumentServiceImpl implements DocumentService {
   public static final String DOCUMENTS_APP_NAVIGATION_NODE_NAME = "documents";
   public static final String DOCUMENT_NOT_FOUND = "?path=doc-not-found";
   private static final String DOCUMENTS_NODE = "Documents";
+  private static final String  ACTIVITY_STREAM_NODE = "Activity Stream Documents";
   private static final String SHARED_NODE = "Shared";
   private static final String COLLABORATION = "collaboration";
   private static final long DEFAULT_EDITORS_IDLE_TIMEOUT = 1800000;
@@ -500,22 +501,14 @@ public class DocumentServiceImpl implements DocumentService {
       if (currentNode.isNodeType(NodetypeConstant.EXO_SYMLINK)) {
         currentNode = linkManager.getTarget(currentNode);
       }
-      List<String> path = new ArrayList<>();
-      Node targetNode = null;
-      boolean existingSymlink = false;
-      for (NodeIterator it = shared.getNodes(); it.hasNext(); ) {
-        Node node = it.nextNode();
-        path.add(((NodeImpl) node).getInternalPath().getAsString());
-        if (path.contains(((NodeImpl) shared).getInternalPath().getAsString() + "[]" + currentNode.getName() + ":1")) {
-          existingSymlink = true;
-          targetNode = node;
-          break;
-        }
+      List<Node> nodes = new ArrayList<Node>();
+      String CurrentNodeWorkspaceName = currentNode.getSession().getWorkspace().getName();
+      nodes = linkManager.getNodeSymlinksUnderFolder(currentNode.getUUID(), shared.getPath(), CurrentNodeWorkspaceName);
+      if (nodes.size() != 0) {
+        link = nodes.get(0);
       }
-      if (existingSymlink) {
-        link = targetNode;
-      } else {
-        link = linkManager.createLink(shared, currentNode);
+      if (link == null && currentNode.isNodeType(NodetypeConstant.NT_FILE)) {
+        link = currentNode;
       }
       return CommonsUtils.getCurrentDomain() + getShortLinkInDocumentsApp(link.getSession().getWorkspace().getName(), ((NodeImpl) link).getInternalIdentifier());
     } catch (Exception e) {
@@ -523,7 +516,7 @@ public class DocumentServiceImpl implements DocumentService {
       return "";
     }
   }
-  
+
   /**
    * {@inheritDoc}
    */
--- a/core/services/src/main/java/org/exoplatform/services/cms/link/LinkManager.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/link/LinkManager.java
@@ -192,5 +192,16 @@ public interface LinkManager {
    * @throws Exception
    */
   public void updateSymlink(Node link) throws Exception;
-  
+
+  /**
+   * get the link of the document inside the folder
+   *
+   * @param contentUUID Jcr uuid of the file
+   * @param folderPath path of the target folder
+   * @param workspace target workspace
+   */
+  default List<Node> getNodeSymlinksUnderFolder(String contentUUID, String folderPath, String workspace) {
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
 }
--- a/core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
@@ -444,4 +444,31 @@ public class LinkManagerImpl implements LinkManager {
     }
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public List<Node> getNodeSymlinksUnderFolder(String contentUUID, String folderPath, String workspace) {
+    List<Node> nodeSymlinks =  new ArrayList<Node>();
+    ManageableRepository repository = WCMCoreUtils.getRepository();
+    SessionProviderService sessionProviderService = CommonsUtils.getService(SessionProviderService.class);
+    SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
+    try {
+      Session session = sessionProvider.getSession(workspace, repository);
+      QueryManager queryManager = session.getWorkspace().getQueryManager();
+      String queryString = new StringBuilder("select * from exo:symlink ").append(" where exo:uuid = '").
+              append(contentUUID).append("' AND jcr:path like '%").
+              append(folderPath).append("/%' order by exo:dateCreated DESC").toString();
+      Query query = queryManager.createQuery(queryString, Query.SQL);
+      QueryResult queryResult = query.execute();
+      NodeIterator iterator = queryResult.getNodes();
+      while (iterator.hasNext()) {
+        nodeSymlinks.add(iterator.nextNode());
+      }
+      return nodeSymlinks;
+    } catch (Exception e) {
+      LOG.error("Error while trying to get the Node Link", e);
+      return null;
+    }
+  }
 }
--- a/ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/listener/ActivityListener.java
+++ b/ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/listener/ActivityListener.java
@@ -35,6 +35,10 @@ public class ActivityListener extends ActivityListenerPlugin {
 
   /** The constant POST_ACTIVITY. */
   private static final Boolean         POST_ACTIVITY = false;
+  /** The constant ACTIVITY PARAMS. */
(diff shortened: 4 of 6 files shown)


core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
```diff
@@ -501,21 +506,25 @@ private String getSharedLink(Node currentNode, Node rootNode) {
...
       }
-      if (existingSymlink) {
-        link = targetNode;
-      } else {
-        link = linkManager.createLink(shared, currentNode);
+      if (!existingSymlink) {
+        for (NodeIterator it = ActivityNode.getNodes(); it.hasNext(); ) {
+          Node node = it.nextNode();
+          path.add(((NodeImpl) node).getInternalPath().getAsString());
+          if (path.contains(((NodeImpl) ActivityNode).getInternalPath().getAsString() + "[]" + currentNode.getName() + ":1")) {
```
ahamdi: What are you trying to check with this test condition ?

ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/listener/ActivityListener.java
```diff
@@ -116,7 +121,7 @@ private void shareActivityFilesToSpace(ActivityLifeCycleEvent activityLifeCycleE
                 }
               }
             } catch (RepositoryException e) {
-              LOG.warn("Error while sharing document to space. Node uuid: " + fileToShare.getId(), e);
+              LOG.warn("Error while sharing document to space. Node uuid: " + nodeUuid, e);
```
ahamdi:
```suggestion
              LOG.error("Error while sharing document to space. Node uuid: " + nodeUuid, e);
```

ahamdi (changes requested):
Please add unit tests for this fix

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
hakermi: If the file is not existing in the shared folder (symlink) so that means its' not a shared file, i could easily say if (link == null) but i let it as it is to be more expressive.

hakermi: If the file (node) was found in the ActivityStreamFolder it will be in this form for example ([]:1[]Groups:1[]spaces:1[]testspace:1[]Documents:1[]Activity Stream Documents:1[]smtp_conf:1;)
so it will be assigned to the link variable and break the loop execution

ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/listener/ActivityListener.java
```diff
@@ -116,7 +121,7 @@ private void shareActivityFilesToSpace(ActivityLifeCycleEvent activityLifeCycleE
                 }
               }
             } catch (RepositoryException e) {
-              LOG.warn("Error while sharing document to space. Node uuid: " + fileToShare.getId(), e);
+              LOG.error("Error while sharing document to space. Node uuid: " + nodeUuid, e);
```
hbenali, line 121: Is it safe to add here (for each iteration) to prevent session leak ?
`            finally {
              sessionProvider.close();
            }
`
cc @ahamdi

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
ahamdi: There are other cleaner ways to do it, let's have a call and I should be able to propose a better way to doit.

hakermi: > There are other cleaner ways to do it, let's have a call and I should be able to propose a better way to doit.

Okay

ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/listener/ActivityListener.java
hakermi: it throws an exception of Kind "sessionprovider already closed"

ahamdi: Yes it is needed to close it otherwise we will have some JCR session  leaks.

core/services/src/main/java/org/exoplatform/services/cms/link/LinkManager.java
```diff
@@ -192,5 +192,17 @@ default boolean isFileOrParentALink(Session session, String path) {
...
+
+  /**
+   * get the link of the document inside the folder
+   *
+   * @param contentUUID Jcr uuid of the file
+   * @param folderPath path of the target folder
+   * @param workspace target workspace
+   * @param nodeType type of current node
+   */
+  default Node getLinkInFolder(String contentUUID, String folderPath, String workspace, String nodeType) {
```
ahamdi: Rename the function and do not add the NodeType inside it. In this service we just manage Symlinks
```suggestion
  default Node getNodeSymlinksUnderFolder(String contentUUID, String folderPath, String workspace) {

```

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
```diff
@@ -444,4 +444,32 @@ public void updateSymlink(Node node) throws RepositoryException {
...
+   */
+  @Override
+  public Node getLinkInFolder(String contentUUID, String folderPath, String workspace, String nodeType) {
+    ManageableRepository repository = WCMCoreUtils.getRepository();
+    SessionProviderService sessionProviderService = CommonsUtils.getService(SessionProviderService.class);
+    SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
+    try {
+      Session session = sessionProvider.getSession(workspace, repository);
+      QueryManager queryManager = session.getWorkspace().getQueryManager();
+      String uuidType = nodeType.equals(NodetypeConstant.EXO_SYMLINK)? NodetypeConstant.EXO_UUID : NodetypeConstant.JCR_UUID;
```
ahamdi: This function is just for getting symlinks. You can not add the logic of checking if a node is a child of another node.

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
```diff
@@ -500,22 +505,10 @@ private String getSharedLink(Node currentNode, Node rootNode) {
...
-        }
-      }
-      if (existingSymlink) {
-        link = targetNode;
-      } else {
-        link = linkManager.createLink(shared, currentNode);
+      String CurrentNodeWorkspaceName = currentNode.getSession().getWorkspace().getName();
+      link = linkManager.getLinkInFolder(currentNode.getUUID(), ((NodeImpl) shared).getInternalPath().getAsString(), CurrentNodeWorkspaceName, EXO_SYMLINK);
+      if (link == null) {
+        link = linkManager.getLinkInFolder(currentNode.getUUID(), ((NodeImpl) activityNode).getInternalPath().getAsString(), CurrentNodeWorkspaceName, NT_FILE);
```
ahamdi: Why ddon't you use Node.getPath() ?

ahamdi (changes requested):

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
hakermi: Node.getPath() returns a path in this form "groups/spaces/doecuments/.../ etc" and it doesn't match the Like query because the path is saved in this form []:1Groups[]:1spaces[]:1Documenst[]:1 etc
The right form to use and to match the like query is ((NodeImpl) activityNode).getInternalPath().getAsString() otherwise the query will not be useful

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
hakermi: What about links of document those are not symlinks ? and not saved in the shared folder ?
If we remove this option the sent link in the emails notification of the activity with attached file will be null and empty

core/services/src/main/java/org/exoplatform/services/cms/documents/impl/DocumentServiceImpl.java
ahamdi: Please recheck again
we always use this form in JCR queries **groups/spaces/doecuments/**

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
ahamdi: > What about links of document those are not symlinks ? and not saved in the shared folder ?
> If we remove this option the sent link in the emails notification of the activity with attached file will be null and empty
You need to cover that case outside the LinkManager

core/services/src/main/java/org/exoplatform/services/cms/link/LinkManager.java
hakermi: done

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
hakermi: done

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
```diff
@@ -444,4 +444,30 @@ public void updateSymlink(Node node) throws RepositoryException {
...
+      Session session = sessionProvider.getSession(workspace, repository);
+      QueryManager queryManager = session.getWorkspace().getQueryManager();
+      String queryString = new StringBuilder("select * from exo:symlink ").append(" where exo:uuid = '").
+              append(contentUUID).append("' AND jcr:path like '%").
+              append(folderPath).append("/%' order by exo:dateCreated DESC").toString();
+      Query query = queryManager.createQuery(queryString, Query.SQL);
+      QueryResult queryResult = query.execute();
+      NodeIterator iterator = queryResult.getNodes();
+      while (iterator.hasNext()) {
+        return iterator.nextNode();
```
ahamdi: We may have multiple symlinks of the same content, let's return a list of nodes

ahamdi (changes requested):

core/services/src/main/java/org/exoplatform/services/cms/link/impl/LinkManagerImpl.java
hakermi: Is it possible to have multiple symlinks of the same content in the same path? why we don't just add LIMIT 1 to the sql query if it's possible ? wdyt?

ahamdi: Yes we may have multiple symlinks inside a folder that targets one document, and of course we can have other symlinks inside child folder of that folder

hakermi: done

ahamdi (approved):
LGTM

PR #1255, opened 2021-05-26
TASK-44558: resize attachment icon in integration app

aelbenney:


apps/portlet-documents/src/main/webapp/vue-app/attachment-integration/components/AttachmentApp.vue
```diff
@@ -49,6 +49,7 @@
                 v-for="attachment in attachmentsToDisplay.slice(0, 2)"
                 :key="attachment.id"
                 :attachment="attachment"
+                :small-attachment-icon="true"
                 :allow-to-remove="false"
                 :allow-to-preview="true" />
```
boubaker:
```suggestion
                :allow-to-remove="false"
                allow-to-preview
                small-attachment-icon />
```

apps/resources-wcm/src/main/webapp/skin/less/ecms/portlets/attachments/attachmentsApp.less
```diff
@@ -20,6 +20,11 @@
     i {
       font-size: 41px;
     }
+    &.smallAttachmentIcon {
+      i {
+        font-size: 24px;
+      }
+    }
```
boubaker:
```suggestion
    &.smallAttachmentIcon i {
      font-size: 24px;
    }
```

aelbenney: it won't work because the `tag i` already has `font-size` [here](https://github.com/exoplatform/ecms/pull/1255/files#diff-cb0bf7f36f5e7277f9351ead8e269b7ec4a8d8d9949831c558b9f04a5507fa78L21)

boubaker: should give exactly the same resulted CSS, no diff about both notations, just less accolades and lines

boubaker (approved):

PR #1260, opened 2021-05-27
TASK-44762: fix displaying removed files in drive explorer

aelbenney:
The displayed files in drive explorer are not updated with the real attached list.

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
@@ -381,6 +381,7 @@ export default {
       files: [],
       space: {},
       currentDrive: {},
+      selectedFiles: [],
       maxFilesCount: 20,
       foldersHistory: [],
       showSearchInput: false,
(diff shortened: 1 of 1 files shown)


boubaker (dismissed):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
```diff
@@ -486,6 +488,9 @@ export default {
     },
     driveExplorerDrawerTitle() {
       return this.modeFolderSelection ? this.$t('attachments.drawer.destination.folder') : this.$t('attachments.drawer.existingUploads');
+    },
+    selectedFiles() {
```
boubaker: we must not change a computed property

boubaker (changes requested):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
boubaker: more explanations: a `computed` property will be automatically refreshed by Vue and is handled by Vue lifecycle. So it's readonly properties. If you which to duplicate a property (`props`) to be able to add changes on it, yiou can add a data for `selectedFiles` + add a watch expression on `attachedFiles` to re-init `selectedFiles`. I hope it's clear enough, is it ?

boubaker (approved):

PR #1277, opened 2021-06-01
TASK-44941: Display different location info for files selected from drive

aelbenney:


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentItem.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div :class="allowToPreview && 'clickable'" class="attachment d-flex">
-    <v-list-item-avatar :class="smallAttachmentIcon ? 'me-0' :'me-3'" class="rounded-lg" @click="openPreview()">
+    <v-list-item-avatar
+      :class="smallAttachmentIcon ? 'me-0' :'me-3'"
+      class="rounded-lg"
```
boubaker: I think we should use helper class `border-radius` as defined in platform-ui

PR #1279, opened 2021-06-01
Task-42406: Fix latest news bloc display when deleting a space having a pinned news

AzmiTouil:
Prior to this change, when we remove a space that has a related pinned news, the bloc latest news disappear. The problem consists of that when deleting the space, the space node is well deleted but not the symlinks of its sub-tree nodes.
To fix this problem, we have deleted all dead symlinks related to the deleted space.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/listeners/NewGroupListener.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/listeners/NewGroupListener.java
@@ -28,6 +28,7 @@ import javax.jcr.Session;
 
 import org.exoplatform.container.ExoContainerContext;
 import org.exoplatform.container.xml.InitParams;
+import org.exoplatform.services.cms.impl.Utils;
 import org.exoplatform.services.jcr.RepositoryService;
 import org.exoplatform.services.jcr.access.PermissionType;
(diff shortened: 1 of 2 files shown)


core/services/src/main/java/org/exoplatform/services/cms/listeners/NewGroupListener.java
```diff
@@ -94,10 +95,16 @@ private void removeGroup(String groupId) throws RepositoryException
       String systemWorkspace = manageableRepository.getConfiguration().getDefaultWorkspaceName();
       Session session = manageableRepository.getSystemSession(systemWorkspace);
       Node groupNode = (Node)session.getItem(groupsPath_ + groupId);
-      groupNode.remove();
-      session.save();
-      session.logout();
-   }
+      try {
+        Utils.removeDeadSymlinks(groupNode);
```
azayati:
```suggestion
        Utils.removeDeadSymlinks(groupNode, false);
```
Removed symlinks should be added to trash but remove definitively

azayati (changes requested):
See review

azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM after QA validation

PR #1289, opened 2021-06-04
TASK-45123: Deleted ActivitiesRestService service

nidhalnidhal:
i deleted ActivitiesRestService class and all its dependencies as it is no longer used to fetch data and it comes with a huge security hole because all end points are not marked with @RolesAllowed


apps/resources-wcm/src/main/webapp/WEB-INF/gatein-resources.xml
```diff
@@ -326,24 +326,6 @@
     </script>
   </module>
 
-  <module>
-    <name>document-preview</name>
```
ahamdi, line 330: Please check that this JS module is not used in GTMPL files

ahamdi (changes requested):

apps/resources-wcm/src/main/webapp/WEB-INF/gatein-resources.xml
nidhalnidhal: after searching , there are 2 gtmpl vues in ecm-wcm-extension using a module named document-preview but ecm-wcm-extension module is configured to use only the right document-preview file which is document-preview.js and not DocumentPreview.js

ahamdi (approved):

apps/resources-wcm/src/main/webapp/WEB-INF/gatein-resources.xml
ahamdi: Ok thanks

PR #1297, opened 2021-06-08
Task-45155 : When uploading a file with combining char in filename, t…

rdenarie:
…he name is not correctly read

Before this fix, when a filename contains combining chararacters, the filename is not correctly interpreted
This fix use normalize js function to remove theses combining characters so that we can interpret the filename


ahamdi (approved):
LGTM

PR #1299, opened 2021-06-08
TASK-44885: Add check icon for selected files in drive explorer

aelbenney:


apps/resources-wcm/src/main/webapp/skin/less/ecms/portlets/attachments/attachmentsApp.less
```diff
@@ -21,6 +21,10 @@
       font-size: 41px;
     }
 
+    .file-type-icon {
+      flex: 0 0 75%;
```
boubaker: please avoid addin new class for this property that you can make with CSS Classes helpers.

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerFileItem.vue
```diff
@@ -5,8 +5,16 @@
       rel="tooltip"
       class="fileTitle d-flex flex-column v-messages"
       data-placement="bottom">
-      <div class="fileType center">
-        <i :class="getIconClassFromFileMimeType(file.mimetype)"></i>
+      <div :class="file.isSelected && 'selected'" class="fileType d-flex">
+        <div class="file-type-icon pa-0 col-sm-9 text-right pt-2">
```
boubaker: `col-sm-9` will not be applied on mobile, you may need to use `col-9` instead

```suggestion
        <div class="file-type-icon pa-0 col-9 text-right pt-2">
```

boubaker (approved):

PR #1305, opened 2021-06-09
MAINT-45398 : Fix the appearing of the context menu in icons view when the name contains double quote chars

hakermi:
**Fix** the appearing of the context menu in icons view when the name contains double quote chars


ahamdi (commented):
Could you add Unit test

hakermi:
> Could you add Unit test

There is already a unit test : https://github.com/exoplatform/ecms/blob/8c24565f771961eaa40717671373fde32532259a/core/connector/src/test/java/org/exoplatform/wcm/connector/collaboration/TestOpenInOfficeConnector.java#L152

hakermi:
The problem is just in the Utils.java class, the changes in the **openInofficeConnector** class has no relation with the fix. it's a just a problem i found it in my way during debug

ahamdi (approved):

PR #1308, opened 2021-06-10
MAINT-45473:Fix the error message appears when edit a short message when it contains a link

hakermi:
**Fix** the error message appears when edit a short message when it contains a link


ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/FileUIActivityBuilder.java
```diff
@@ -77,7 +77,7 @@ protected void extendUIActivity(BaseUIActivity uiActivity, ExoSocialActivity act
     // So we couldn't set activity.getTitle() all the time, see INTEG-486
     if (activity.getTemplateParams() != null
         && StringUtils.isNotBlank(templateParams.get(FileUIActivity.MESSAGE))) {
-      fileActivity.setMessage(templateParams.get(FileUIActivity.MESSAGE));
+      fileActivity.setMessage(templateParams.get(FileUIActivity.MESSAGE).split("<oembed>")[0]);
```
ahamdi, line 80: You need check if the message contains that tag before getting the first element otherwise you will get an ArrayOutOfBoundsException

ecms-social-integration/src/test/java/org/exoplatform/wcm/ext/component/activity/FileUIActivityTest.java
```diff
@@ -42,16 +42,15 @@ public void testActivityMessageToDisplay() throws Exception {
     activity.setTitle(activityTitle);
 
     Map<String, String> activityParameters = new HashMap<>();
-    activityParameters.put(FileUIActivity.MESSAGE, "message test");
+    activityParameters.put(FileUIActivity.MESSAGE, "message test<oembed>link</oembed>");
```
ahamdi, line 45: add the check for a message that does not contain the tag <oembed>

ahamdi (changes requested):

ahamdi (approved):

PR #1335, opened 2021-06-30
MAINT-45398: Fix the issue of opening a document in the content explorer when its name contains double quotes char

hakermi:
**Fix** the issue of opening a document in the content explorer when its name contains double quotes char by cleaning the node physical name when renaming the document.

**Note** : this may affect temporary the old documents which their node names contains double quotes : renaming them again will fix the problem.

As it's not possible to escape the double quotes in the html attributes values, without this fix we will have always an issue opening the document or displaying the context menu by right click when the name of document contains a double quote char.


core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/RenameConnector.java
```diff
@@ -113,6 +113,10 @@ public Response rename(@QueryParam("oldPath") String oldPath,
       // Clarify new name & check to add extension
 
       String newName  = Text.escapeIllegalJcrChars(newTitle);
+
+      //clean node name
+      newName = Utils.cleanName(newName);
```
ahamdi, line 118: Ok this will help renaming contents, whet should we do with other documents having quotes in their names ?

hakermi: For uploading docs this is already handled : 
https://github.com/exoplatform/ecms/blob/7e780b752e8798c2a747e2836f9631bf438780d1/core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java#L417

for the affected renamed doc before this fix, we need only to rename them again, I already merged a PR to allow displaying the context menu for docs with double quote char. https://github.com/exoplatform/ecms/pull/1305

ahamdi (approved):

PR #1336, opened 2021-07-01
TASK-45531: Implement change location for attachments

aelbenney:


core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -282,6 +301,40 @@ public Attachment getAttachmentById(String attachmentId, SessionProvider session
...
+      throw new IllegalAccessException("User identity must be positive");
+    }
+    if (StringUtils.isEmpty(attachmentId)) {
+      throw new IllegalArgumentException("Attachment id is mandatory");
+    }
+    if (StringUtils.isEmpty(newPathDrive)) {
+      throw new IllegalArgumentException("New destination path's drive must not be empty");
+    }
+
+    SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
```
boubaker: Please avoid using system session, rather use user session for permission checks

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -282,6 +301,40 @@ public Attachment getAttachmentById(String attachmentId, SessionProvider session
...
+    String workspace = repositoryService.getCurrentRepository().getConfiguration().getDefaultWorkspaceName();
+    Session session = sessionProvider.getSession(workspace, repositoryService.getCurrentRepository());
+    try {
+      Node attachmentNode = session.getNodeByUUID(attachmentId);
+      Node destNode = getParentFolderNode(workspace, session, newPathDrive, newPath);
+      String destPath = destNode.getPath().concat("/").concat(attachmentNode.getName());
+      session.move(attachmentNode.getPath(), destPath);
+      attachmentNode.refresh(true);
```
boubaker: why refresing, isn't the `session.save()` sufficient ?

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -282,6 +301,40 @@ public Attachment getAttachmentById(String attachmentId, SessionProvider session
...
+      session.save();
+    } catch (Exception e) {
+      LOG.error("Error while trying to move attachment node with id {} to the new path {} " + attachmentId, newPath, e);
```
boubaker: please avoid swallowing exception and throw the exception to REST endpoint that will handle it to return the adequate HTTP status code to end user. (No need to log here neither)

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +360,42 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+   * @param currentFolder the current folder
+   *
+   * @return the parent folder node
+   *
+   * @throws Exception the exception
+   */
+  private Node getParentFolderNode(String workspaceName, Session session, String driverName, String currentFolder) throws Exception {
+    DriveData driveData = manageDriveService.getDriveByName(driverName);
+    StringBuilder parentPath = new StringBuilder("");
+    if(driveData != null) {
```
boubaker: formatting

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +360,42 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+      parentPath.append(driveData.getHomePath());
+    }
+    NodeHierarchyCreator nodeHierarchyCreator = WCMCoreUtils.getService(NodeHierarchyCreator.class);
```
boubaker: Use `ExoContainerContext.getService` instead. Why not adding this service as dependency declared in constructor ?

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +360,42 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+
+    if (driveData != null
```
boubaker: merge this condition with previous line:
    if(driveData != null) {
    if(driveData != null ...

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +360,42 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+      parentPath.append(Utils.getPersonalDrivePath(driveData.getHomePath(), ConversationState.getCurrent().getIdentity().getUserId()));
+    }
+
+    parentPath.append((currentFolder != null && currentFolder.length() != 0) ? "/" : "").append(currentFolder);
+    String parentPathStr = parentPath.toString().replace("//", "/");
+    return getTargetNode(session, parentPathStr);
+  }
+
+  private Node getTargetNode(Session session, String path) throws Exception {
+    NodeFinder nodeFinder = WCMCoreUtils.getService(NodeFinder.class);
```
boubaker: To add as constructor dependency

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +360,42 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+    Node node = (Node) nodeFinder.getItem(session, path, true);
+    return node;
```
boubaker:
```suggestion
    return (Node) nodeFinder.getItem(session, path, true);
```

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -258,6 +258,37 @@ public Response deleteEntityAttachment(@ApiParam(value = "Entity technical ident
     }
   }
 
+  @POST
+  @Path("/{attachmentId}/move")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed("users")
+  @ApiOperation(value = "move an attachment to a destination path", httpMethod = "DELETE", response = Response.class, consumes = "application/json", notes = "returns empty response")
```
boubaker:
```suggestion
  @ApiOperation(value = "Move an attachment to a destination path", httpMethod = "POST", response = Response.class, consumes = "application/json", notes = "returns empty response")
```

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -258,6 +258,37 @@ public Response deleteEntityAttachment(@ApiParam(value = "Entity technical ident
...
+    }
+
+    long userIdentityId = getCurrentUserIdentityId();
+    try {
+      attachmentService.moveAttachmentToNewPath(userIdentityId, attachmentId, newPathDrive, newPath);
+    } catch (Exception e) {
+      LOG.error("Error when trying to move attachment with id {} to new destination path {} ", attachmentId, newPath, e);
+      return Response.serverError().entity(e.getMessage()).build();
+    }
+    return Response.ok().build();
```
boubaker:
```suggestion
    return Response.noContent().build();
```

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
aelbenney: okay

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
aelbenney: yes it is sufficient.

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
aelbenney: missed it when refactoring the old code.

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
aelbenney: okay

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
aelbenney: yesss

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -258,6 +258,37 @@ public Response deleteEntityAttachment(@ApiParam(value = "Entity technical ident
     }
   }
 
+  @POST
+  @Path("/{attachmentId}/move")
+  @Produces(MediaType.APPLICATION_JSON)
```
boubaker:
```suggestion
  @Consumes(MediaType.APPLICATION_JSON)
```

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -307,4 +371,38 @@ private void sortAttachments(List<AttachmentContextEntity> attachments) {
...
+    if(driveData != null) {
+      parentPath.append(driveData.getHomePath());
+
+      if (driveData.getHomePath().startsWith(nodeHierarchyCreator.getJcrPath(BasePath.CMS_USERS_PATH) + "/${userId}")) {
+        parentPath.setLength(0);
+        parentPath.append(Utils.getPersonalDrivePath(driveData.getHomePath(), ConversationState.getCurrent().getIdentity().getUserId()));
+      }
+    }
+    
+    parentPath.append((currentFolder != null && currentFolder.length() != 0) ? "/" : "").append(currentFolder);
```
boubaker:
```suggestion
    if (StringUtils.isNotBlank(currentFolder)) {
      parentPath.append("/").append(currentFolder);
    }
```

apps/portlet-documents/src/main/webapp/js/attachmentService.js
```diff
@@ -271,7 +266,34 @@ export function getAttachmentById(attachmentId) {
...
+  }
+  if (entityType) {
+    params.entityType = entityType;
+  }
+  if (entityId) {
+    params.entityId = entityId;
+  }
+  params = $.param(params, true);
+
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/attachments/${attachmentId}/move?${params}`, {
```
boubaker: It would be better to pass the parameters in Body instead of query (Possible Seurity breach, where the names are in the URL and thus not encrypted with TLS/SSL)

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -184,16 +186,25 @@ public Response getAttachmentsByEntity(@ApiParam(value = "Entity technical ident
...
+    }
+    long userIdentityId = getCurrentUserIdentityId();
     try {
-      Attachment attachment = attachmentService.getAttachmentById(attachmentId, WCMCoreUtils.getUserSessionProvider());
+      Attachment attachment = attachmentService.getAttachmentById(entityType, entityId, attachmentId, userIdentityId);
       if (attachment == null) {
         return Response.status(Status.NOT_FOUND).build();
       } else {
-        return Response.ok(EntityBuilder.fromAttachment(identityManager, attachment)).build();
+        return Response.ok(attachment).build();
```
boubaker: you should return the DTO of REST instead of the one used for Service layer

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -239,23 +266,24 @@ public void deleteAttachmentItemById(long userIdentityId,
     sortAttachments(attachmentsContextEntity);
     List<Attachment> attachments = new ArrayList<>();
     if (!attachmentsContextEntity.isEmpty()) {
-      SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
-      String workspace = repositoryService.getCurrentRepository().getConfiguration().getDefaultWorkspaceName();
-      Session session = sessionProvider.getSession(workspace, repositoryService.getCurrentRepository());
+      Session session = getSession();
       try {
         for (AttachmentContextEntity attachmentContextEntity : attachmentsContextEntity) {
```
boubaker: No JPA entity in Service layer please

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -302,9 +380,59 @@ public boolean canDelete(long userIdentityId, String entityType, long entityId)
...
 
+  public boolean canEdit(long userIdentityId, String entityType, long entityId) {
+    AttachmentACLPlugin attachmentACLPlugin = this.aclPlugins.get(entityType);
+    if (attachmentACLPlugin == null) {
+      return false;
+    }
+    return attachmentACLPlugin.canEdit(userIdentityId, entityType, String.valueOf(entityId));
+  }
+
   private void sortAttachments(List<AttachmentContextEntity> attachments) {
```
boubaker: No JPA entity in service layer please

core/services/src/main/java/org/exoplatform/services/attachments/storage/AttachmentStorageImpl.java
```diff
@@ -32,16 +32,13 @@ public AttachmentStorageImpl(AttachmentDAO attachmentDAO) {
   }
 
   @Override
-  public void linkAttachmentsToEntity(long entityId, String entityType, List<String> attachmentsIds) {
-    attachmentsIds.forEach(attachmentId -> {
+  public void linkAttachmentToEntity(long entityId, String entityType, String attachmentId) {
       AttachmentContextEntity attachmentContextEntity = new AttachmentContextEntity();
```
boubaker: Please avoid handling JPA entities in service layer

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -211,14 +250,18 @@ public void deleteAttachmentItemById(long userIdentityId,
...
 
-    AttachmentContextEntity attachmentContextEntity = attachmentStorage.getAttachmentItemByEntity(entityId,
-                                                                                                  entityType,
-                                                                                                  attachmentId);
-    if (attachmentContextEntity == null) {
+    Attachment attachmentEntity = null;
+    try {
+      attachmentEntity = attachmentStorage.getAttachmentItemByEntity(entityId, entityType, attachmentId);
+    } catch (Exception e) {
+      throw new NotFoundException("Can't convert attachment JCR node to Attachment DTO", e);
```
boubaker: why using Clouddrive exception ?

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -19,20 +19,31 @@
...
 import org.exoplatform.commons.exception.ObjectNotFoundException;
-import org.exoplatform.services.attachments.model.*;
+import org.exoplatform.services.attachments.model.Permission;
+import org.exoplatform.services.attachments.model.Attachment;
 import org.exoplatform.services.attachments.plugin.AttachmentACLPlugin;
 import org.exoplatform.services.attachments.storage.AttachmentStorage;
 import org.exoplatform.services.attachments.utils.EntityBuilder;
+import org.exoplatform.services.attachments.utils.Utils;
+import org.exoplatform.services.cms.BasePath;
+import org.exoplatform.services.cms.clouddrives.NotFoundException;
```
boubaker: To replace by a more generic checked exception

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -122,11 +146,14 @@ public void updateEntityAttachments(long userIdentityId,
...
-    List<AttachmentContextEntity> existingAttachmentsContext =
-                                                             attachmentStorage.getAttachmentContextByEntity(entityId, entityType);
-    List<String> existingAttachmentsIds = existingAttachmentsContext.stream()
-                                                                    .map(AttachmentContextEntity::getAttachmentId)
-                                                                    .collect(Collectors.toList());
+    List<Attachment> existingEntityAttachments = null;
+    try {
+      existingEntityAttachments = attachmentStorage.getAttachmentsByEntity(userIdentityId, entityId, entityType);
+    } catch (NotFoundException | RepositoryException e) {
+      throw new ItemNotFoundException("Error while trying to get attachments of the entity with type" + entityType + "and id"
```
boubaker: Throwing a checked exception of JCR storage layer isn't convenient here IMO. You can define your own checked exceptions if needed

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -169,11 +203,14 @@ public void deleteAllEntityAttachments(long userIdentityId, long entityId, Strin
...
-    List<AttachmentContextEntity> existingAttachmentsContext =
-                                                             attachmentStorage.getAttachmentContextByEntity(entityId, entityType);
-    List<String> attachmentsIds = existingAttachmentsContext.stream()
-                                                            .map(AttachmentContextEntity::getAttachmentId)
-                                                            .collect(Collectors.toList());
+    List<Attachment> existingEntityAttachments = null;
+    try {
+      existingEntityAttachments = attachmentStorage.getAttachmentsByEntity(userIdentityId, entityId, entityType);
+    } catch (NotFoundException | RepositoryException e) {
+      throw new ItemNotFoundException("Error while trying to get attachments of the entity with type" + entityType + "and id"
```
boubaker: Same here as https://github.com/exoplatform/ecms/pull/1336/files#r664592410

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
boubaker: I suggest to avoid catching the exception here and let upper layer receive and handle the exception, unless you have an important information to include in exception that will help, diagnose the original problem on runtime (which is not the case, because the message is a static string input and no variables or additional information is added)

core/services/src/test/java/org/exoplatform/services/attachments/service/AttachmentServiceTest.java
```diff
@@ -100,25 +127,21 @@ public void testLinkAttachmentsToEntity() throws Exception { // NOSONAR
...
+      attachmentService.linkAttachmentToEntity(0, 5, "task", "1");
       fail();
     } catch (IllegalAccessException e) {
       // Expected
     }
 
-    //when
-    when(sessionProviderService.getSystemSessionProvider(any())).thenReturn(sessionProvider);
+    // when
+    lenient().when(sessionProviderService.getSystemSessionProvider(any())).thenReturn(sessionProvider);
```
boubaker: are we using system session somewhere in the code ?

apps/portlet-documents/src/main/webapp/js/attachmentService.js
```diff
@@ -189,24 +189,23 @@ export function uploadAttachment(workspaceName, driveName, currentFolder, curren
...
-  params = $.param(params, true);
 
-  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/attachments/${entityType}/${entityId}?${params}`, {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/attachments/${entityType}/${entityId}`, {
     credentials: 'include',
     method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: attachmentId,
```
boubaker: if it's only about one attachmentId, it would be better to add it in Path param in this case

apps/portlet-documents/src/main/webapp/js/attachmentService.js
```diff
@@ -271,7 +270,34 @@ export function getAttachmentById(attachmentId) {
...
+  if (newPathDrive) {
+    params.newPathDrive = newPathDrive;
+  }
+  if (entityType) {
+    params.entityType = entityType;
+  }
+  if (entityId) {
+    params.entityId = entityId;
+  }
+  params = $.param(params, true);
```
boubaker: please avoid using jQuery when possible and use FormData javascript object

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentService.java
```diff
@@ -18,22 +18,32 @@
...
-  void deleteAllEntityAttachments(long userIdentityId, long entityId, String entityType) throws ObjectNotFoundException, IllegalAccessException;
+  void deleteAllEntityAttachments(long userIdentityId, long entityId, String entityType) throws ObjectNotFoundException, IllegalAccessException ;
 
-  void deleteAttachmentItemById(long userIdentityId, long entityId, String entityType, String attachmentId) throws ObjectNotFoundException, IllegalAccessException;
+  void deleteAttachmentItemById(long userIdentityId,
+                                long entityId,
+                                String entityType,
+                                String attachmentId) throws ObjectNotFoundException, IllegalAccessException ;
 
+  void moveAttachmentToNewPath(long userIdentityId,
```
boubaker, line 136: need javadoc on API methods

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -169,11 +187,14 @@ public void deleteAllEntityAttachments(long userIdentityId, long entityId, Strin
...
-    List<AttachmentContextEntity> existingAttachmentsContext =
-                                                             attachmentStorage.getAttachmentContextByEntity(entityId, entityType);
-    List<String> attachmentsIds = existingAttachmentsContext.stream()
-                                                            .map(AttachmentContextEntity::getAttachmentId)
-                                                            .collect(Collectors.toList());
+    List<Attachment> existingEntityAttachments = null;
+    try {
+      existingEntityAttachments = attachmentStorage.getAttachmentsByEntity(userIdentityId, entityId, entityType);
+    } catch (ObjectNotFoundException e) {
+      throw new ObjectNotFoundException("Error while trying to get attachments of the entity with type" + entityType + "and id"
```
boubaker: it's useless to catch and throw the same exception just after

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -142,15 +156,19 @@ public void updateEntityAttachments(long userIdentityId,
     if (attachmentIds != null && !attachmentIds.isEmpty()) {
       for (String attachmentId : attachmentIds) {
         if (!existingAttachmentsIds.contains(attachmentId) && StringUtils.isNotEmpty(attachmentId)) {
-          linkAttachmentsToEntity(userIdentityId, entityId, entityType, Collections.singletonList(attachmentId));
+          try {
+            linkAttachmentToEntity(userIdentityId, entityId, entityType, attachmentId);
+          } catch (Exception e) {
+            throw new ObjectNotFoundException("Can't link attachment with id " + attachmentId + " to entity type " + entityType
```
boubaker: the thrown exception here is not always saying the truth IMO. (you throw ObjectNotFoundException for any type of exception, even, for example, session closed...)

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -211,18 +232,22 @@ public void deleteAttachmentItemById(long userIdentityId,
...
 
-    AttachmentContextEntity attachmentContextEntity = attachmentStorage.getAttachmentItemByEntity(entityId,
-                                                                                                  entityType,
-                                                                                                  attachmentId);
-    if (attachmentContextEntity == null) {
+    Attachment attachmentEntity = null;
+    try {
+      attachmentEntity = attachmentStorage.getAttachmentItemByEntity(entityId, entityType, attachmentId);
+    } catch (ObjectNotFoundException e) {
+      throw new ObjectNotFoundException("Can't convert attachment JCR node to Attachment DTO");
```
boubaker: no need to catch the original exception to rethrow it just after

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -235,46 +260,86 @@ public void deleteAttachmentItemById(long userIdentityId,
...
+                                                    attachmentId);
+
+      boolean canView = canView(userIdentityId, entityType, entityId);
+      boolean canDelete = canDelete(userIdentityId, entityType, entityId);
+      boolean canEdit = canEdit(userIdentityId, entityType, entityId);
+      Permission attachmentACL = new Permission(canView, canDelete, canEdit);
+
+      attachment.setAcl(attachmentACL);
+    } catch (Exception e) {
+      throw new ObjectNotFoundException("Can't convert attachment JCR node with id " + attachmentId + " to entity");
```
boubaker: not compatible exceptions, can't transform all `Exception` => `ObjectNotFoundException`

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -235,46 +260,86 @@ public void deleteAttachmentItemById(long userIdentityId,
...
   @Override
-  public Attachment getAttachmentById(String attachmentId, SessionProvider sessionProvider) throws Exception {
+  public void moveAttachmentToNewPath(long userIdentityId,
+                                      String attachmentId,
+                                      String newPathDrive,
+                                      String newPath,
+                                      String entityType,
+                                      long entityId) throws ObjectNotFoundException, IllegalAccessException {
+    if (userIdentityId <= 0) {
+      throw new IllegalAccessException("User identity must be positive");
```
boubaker: `IllegalAccessException` => `IllegalArgumentException`

core/services/src/main/java/org/exoplatform/services/attachments/service/AttachmentServiceImpl.java
```diff
@@ -235,46 +260,86 @@ public void deleteAttachmentItemById(long userIdentityId,
...
+                                                manageDriveService,
+                                                nodeHierarchyCreator,
+                                                nodeFinder,
+                                                newPathDrive,
+                                                newPath);
+      String destPath = destNode.getPath().concat("/").concat(attachmentNode.getName());
+      session.move(attachmentNode.getPath(), destPath);
+      session.save();
+    } catch (Exception e) {
+      throw new ObjectNotFoundException("Error while trying to move attachment node with id " + attachmentId + " to the new path "
```
boubaker: not compatible exceptions, can't transform all `Exception` => `ObjectNotFoundException`

core/services/src/main/java/org/exoplatform/services/attachments/storage/AttachmentStorageImpl.java
```diff
@@ -16,42 +16,112 @@
...
+          Attachment attachment = EntityBuilder.fromAttachmentNode(repositoryService,
+                                                                   documentService,
+                                                                   linkManager,
+                                                                   Utils.getCurrentWorkspace(repositoryService),
+                                                                   session,
+                                                                   attachmentContextEntity.getAttachmentId());
+          attachments.add(attachment);
+        }
+      } catch (Exception e) {
+        throw new ObjectNotFoundException("Can't convert attachment JCR node to Attachment DTO");
```
boubaker: not compatible exceptions, can't transform all `Exception` => `ObjectNotFoundException`

core/services/src/main/java/org/exoplatform/services/attachments/storage/AttachmentStorageImpl.java
```diff
@@ -16,42 +16,112 @@
...
+    try {
+      session = Utils.getSession(sessionProviderService, repositoryService);
+      attachment = EntityBuilder.fromAttachmentNode(repositoryService,
+                                                    documentService,
+                                                    linkManager,
+                                                    Utils.getCurrentWorkspace(repositoryService),
+                                                    session,
+                                                    attachmentEntity.getAttachmentId());
+    } catch (Exception e) {
+      throw new ObjectNotFoundException("Can't convert attachment JCR node to Attachment DTO");
```
boubaker: not compatible exceptions, can't transform all `Exception` => `ObjectNotFoundException`

core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -302,6 +301,39 @@ public Response downloadActivityAttachments(
...
+    if (StringUtils.isEmpty(attachmentId)) {
+      return Response.status(Response.Status.BAD_REQUEST).entity("Attachment id is mandatory").build();
+    }
+
+    long userIdentityId = getCurrentUserIdentityId();
+    try {
+      attachmentService.moveAttachmentToNewPath(userIdentityId, attachmentId, newPathDrive, newPath, entityType, entityId);
+    } catch (Exception e) {
+      LOG.error("Error when trying to move attachment with id {} to new destination path {} ", attachmentId, newPath, e);
+      return Response.serverError().entity(e.getMessage()).build();
```
boubaker: it's better to avoid giving the user access to the exception message which can hold a sensitive data for logging only.
You should add a simple message that can helps the user to change his query, or nothing when it's about a generic exception as here.

boubaker (approved):

PR #1337, opened 2021-07-01
TASK-45883: fixed @mention disfunction in activitystream

nidhalnidhal:
Before this fix mentioning was not functional in activity stream when the post contains an attachement , so i fixed this behavior by rendring the exact formatted content of the post.


ecms-social-integration/src/main/java/org/exoplatform/wcm/ext/component/activity/FileUIActivity.java
```diff
@@ -767,7 +767,7 @@ public String getActivityStatus() {
     if (message == null) {
```
ahamdi: Why do we check on the message and return the title of the activity ?

ahamdi (approved):

PR #1345, opened 2021-07-06
MAINT-45398 : Fix the appearing of the context menu in icons view when the name contains double quote chars

mzorai:
**Fix** the appearing of the context menu in icons view when the name contains double quote chars


hbenali (approved):
LGTM, good catch (y) :)

PR #1351, opened 2021-07-12
TASK-45532: Rework removing attached files to an entity

aelbenney:


core/services/src/main/java/org/exoplatform/services/rest/AttachmentsRestService.java
```diff
@@ -273,7 +273,7 @@ public Response deleteAttachmentsByEntity(@ApiParam(value = "Entity technical id
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed("users")
   @ApiOperation(value = "Delete an attachment linked to the given entity", httpMethod = "DELETE", response = Response.class, consumes = "application/json", notes = "returns empty response")
```
boubaker:
```suggestion
  @ApiOperation(value = "Delete an attachment linked to the given entity", httpMethod = "DELETE", response = Response.class, consumes = "application/json", notes = "returns deleted attachment")
```

PR #1366, opened 2021-07-29
TASK-46918: Decode Attachment title

aelbenney:
The attachment title saved in DB is encoded. We need to unescape the title in order to well display it.


ahamdi (approved):
LGTM

PR #1372, opened 2021-08-03
Task-43174 : Blank page received once sharing a document

ahmadmido:
This PR will fix the problem of receiving a blank page after sharing a document with a user on space. The content of the page is disparated, we need to reload the page to have the next context.


apps/resources-wcm/src/main/webapp/javascript/eXo/wcm/frontoffice/public/ShareContent.js
```diff
@@ -123,6 +123,7 @@
       gj(".PopupContent .uiActionBorder .btn-primary").attr("disabled","disabled");
       gj("#shareActionBtn").trigger("click");
     });
+    location.reload(true);
```
ahamdi, line 126: This is a workaround, a popup should be displayed indicating that the file was shared without loosing the File Explorer application in background. Please check the action in webUI and see why the file explorer disappears after sharing the document.

ahamdi (changes requested):

PR #1390, opened 2021-08-11
Task-46769 : Image disappearing after a reset server on notes (#1387)

mkrout:
* Task-46769 : Image disappearing after a reset server on notes


azayati (approved):
LGTM

PR #1392, opened 2021-08-12
TASK-42087: fixed spaces accesibility when creating a symilink from drives application

nidhalnidhal:
Before this fix , when creating a symilink from drives application , an administrator has full access to application spaces even the ones which he is not a member of . So i fixed this by filtering the tree entries when selecting "spaces" from the symilink creation popup in order to keep only the spaces where the administrator is a member.

### DIFF
--- a/core/webui/src/main/java/org/exoplatform/ecm/webui/tree/UINodeTreeBuilder.java
+++ b/core/webui/src/main/java/org/exoplatform/ecm/webui/tree/UINodeTreeBuilder.java
@@ -33,7 +33,10 @@ import org.exoplatform.ecm.webui.utils.Utils;
 import org.exoplatform.portal.webui.container.UIContainer;
 import org.exoplatform.services.cms.link.NodeFinder;
 import org.exoplatform.services.ecm.publication.PublicationService;
+import org.exoplatform.services.security.ConversationState;
 import org.exoplatform.services.wcm.core.NodeLocation;
+import org.exoplatform.social.core.space.model.Space;
+import org.exoplatform.social.core.space.spi.SpaceService;
 import org.exoplatform.webui.application.WebuiRequestContext;
 import org.exoplatform.webui.config.annotation.ComponentConfig;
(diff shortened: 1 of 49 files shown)


rdenarie (changes requested):
We should wait to continue this one.
I'm not sure it is a bug

core/webui/src/main/java/org/exoplatform/ecm/webui/tree/UINodeTreeBuilder.java
```diff
@@ -238,14 +243,19 @@ private void addNodePublish(List<Node> listNode, Node node, PublicationService p
...
     List<Node> list = new ArrayList<Node>();
     if (acceptedNodeTypes.length > 0) {
       for(;iterator.hasNext();) {
         Node sibbling = iterator.nextNode();
         if(sibbling.isNodeType("exo:hiddenable")) continue;
         for(String nodetype: acceptedNodeTypes) {
           if(sibbling.isNodeType(nodetype)) {
-            list.add(sibbling);
+            Space space = spaceService.getSpaceByDisplayName(sibbling.getName());
+           if ((space == null) || (space!=null) && (spaceService.isMember(space,currentUser))) {
```
ahamdi, line 256:
```suggestion
           if ((space == null) || (space != null && spaceService.isMember(space,currentUser))) {

```

core/webui/src/main/java/org/exoplatform/ecm/webui/tree/UINodeTreeBuilder.java
```diff
@@ -238,14 +243,19 @@ private void addNodePublish(List<Node> listNode, Node node, PublicationService p
   }
 
   private List<Node> filfer(final NodeIterator iterator) throws Exception{
```
ahamdi, line 245:
```suggestion
  private List<Node> filter(final NodeIterator iterator) throws Exception{

```

ahamdi (changes requested):

PR #1396, opened 2021-08-17
TASK-43174: Create a new system session provider to avoid using it after being closed

aelbenney:
The problem was that the `UIJCRExplorerPortlet` was not able to get the current node to refresh the page because the session was closed.
This fix creates a new system session provider to be always sure we can instantiate a new session.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
@@ -306,7 +306,7 @@ public class Utils {
   }
 
   public static String getPersonalDrivePath(String parameterizedDrivePath, String userId) throws Exception {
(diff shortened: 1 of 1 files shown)


core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
```diff
@@ -306,7 +306,7 @@ private static String getVersionHistory(String valueHistory) {
   }
 
   public static String getPersonalDrivePath(String parameterizedDrivePath, String userId) throws Exception {
-    SessionProvider sessionProvider = WCMCoreUtils.getSystemSessionProvider();
+    SessionProvider sessionProvider = SessionProvider.createSystemProvider();
```
ahamdi, line 309: Could you close the SessionProvider ?

ahamdi (approved):

PR #1411, opened 2021-08-23
MAINT-48141: Fix Expand folder contents via the expand icon button in the left side explorer TreeNode when it's paginated

hakermi:
**ISSUE** : When a folder has more then 20 child sub nodes it doesn't being expanded when click on expand icon because when building the pageList the expand option is not set for the TreeNode..
**FIX** : Set the Expand option when building the pageList for the targeted TreeNode


rdenarie (approved):

rdenarie (approved):
Dont forget to put your PR description as commit message when squashing

PR #1417, opened 2021-08-25
Task 48159 : Warning popup UI is badly displayed

Rdidmohamedelhadi:
**Problem:** we use two keys that are not added to the properties file.
 **the solution**: I added these keys `attachments.filesFoldersSelector.popup.button.ok`  `attachments.filesFoldersSelector.action.delete.popup.button.cancel` to file properties


rdenarie (commented):
Code is ok, but waiting for a PO precision on task

Rdidmohamedelhadi:
> Code is ok, but waiting for a PO precision on task

The **ok** (`attachments.filesFoldersSelector.popup.button.ok`) key is displayed in the warning popup when you already have the folder name. The **cancel** (`attachments.filesFoldersSelector.action.delete.popup.button.cancel`) key is displayed when you delete a folder, so in this case, you need two buttons (**ok**, **cancel**) !! cc @mzorai

rdenarie (commented):
Can you update the commit message, to follow the format 
<Title>
<Empty line>
<Description> (answering WHY ? and HOW ?)

Thanks

rdenarie (approved):
Approved

Next time, can you try to use imperative form (This change modifies ...) instead of using I

Thanks

Rdidmohamedelhadi:
> Next time, can you try to use imperative form (This change modifies ...) instead of using I

ok !

PR #1418, opened 2021-08-26
Task-47909: Fix display of destination path.

mdiha:
The destination path of attachments in new composer is badly displayed.
This change removes the unneeded css property responsible for wrongly displaying the destination path.


rdenarie (changes requested):
Can you put the PR description in the commit message ? so that we do not forget it when squashing the PR

mdiha:
> Can you put the PR description in the commit message ? so that we do not forget it when squashing the PR

Done

rdenarie (approved):

PR #1419, opened 2021-08-26
MAINT-28272: Fix snapshot myWork documents portlet listing order update

hakermi:
Additional PR for https://github.com/Meeds-io/commons/pull/232
**ISSUE**: The elasticsearch query for retieving documents wasn't musting the modifiedBy field to be only the user who modified the document, 
So, when someone else updates the documents which im the owner, it goes on up the list of myWork docs as that i'm the modifier while it's not the case.
**FIX**: Correct the elasticSearch filter by using the bool query must cluase instead of should in this particular case.


rdenarie:
As the PR is in draft, I let you work on it.
Please mention me when it is ok for review

hakermi:
> As the PR is in draft, I let you work on it.
> Please mention me when it is ok for review

@rdenarie ready for review

rdenarie (changes requested):
Can you put the PR description in the first commit message

hakermi:
> Can you put the PR description in the first commit message

Done

rdenarie (approved):

PR #1424, opened 2021-08-30
Task-48393 : Folders drive is badly displayed in Attach files drawer

AzmiTouil:
Prio this change , the drive folders in the attach files drawer is badly displayed with an extra space between drive icon and title.
Fix : remove extra spacing.


rdenarie (changes requested):
Can  you rebase your PR in order to have only your commit ?

In addition can you complete your commit message ? You only answer the WHY ? question, but the HOW is missing.

Thanks

rdenarie (approved):

PR #1425, opened 2021-09-01
Task-48221 : Download option is still displayed after suspend temporarily download documents on transfer rules page.

ahmadmido:
Issue: Before this fix, when suspending temporarily download documents on the transfer rules page, the download action (download or downloadAll)is still displayed on the activity stream (3 dots button).
Solution: Check the transfer rules download action status and filter enabled action according to this value, if suspend is enabled the download actions are displayed(download or downloadAll) otherwise isn't displayed.

### DIFF
--- a/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
+++ b/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
@@ -1,0 +1,13 @@
+export function getTransfertRulesDownloadDocumentStatus() {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/transferRules/getTransfertRulesDocumentStatus`, {
+    method: 'GET',
+    credentials: 'include',
+  }).then((resp) => {
+    if (resp && resp.ok) {
+      return resp.json();
+    }
+    else {
+      throw new Error ('Error when getting transfer rules download documents status');
+    }
+  });
+}
--- a/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
+++ b/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
@@ -9,5 +9,10 @@ export function getTransfertRulesDownloadDocumentStatus() {
     else {
       throw new Error ('Error when getting transfer rules download documents status');
     }
+  }).then(data => {
+    if (data) {
+      const downloadStatusPromise =  data.downloadDocumentStatus !== 'true';
+      return Promise.resolve(downloadStatusPromise);
+    }
   });
 }
--- a/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
+++ b/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
@@ -11,8 +11,7 @@ export function getTransfertRulesDownloadDocumentStatus() {
     }
   }).then(data => {
     if (data) {
-      const downloadStatusPromise =  data.downloadDocumentStatus !== 'true';
-      return Promise.resolve(downloadStatusPromise);
+      return data.downloadDocumentStatus !== 'true';
     }
   });
 }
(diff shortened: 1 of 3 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/extensions.js
```diff
@@ -54,30 +54,35 @@ export function installExtensions() {
...
-    isEnabled: activity => {
-      if (activity.templateParams) {
-        const docPaths = activity.templateParams.DOCPATH && activity.templateParams.DOCPATH.split('|@|')
-                          || (activity.templateParams.nodePath && [activity.templateParams.nodePath]);
-        return docPaths && docPaths.length === 1;
-      }
-    },
-    rank: 0,
-  }, downloadHandlerExtension));
+  Vue.prototype.$transferRulesService.getTransfertRulesDownloadDocumentStatus()
```
boubaker, line 57: IMO, it's better to make the method `getTransfertRulesDownloadDocumentStatus` returns a Promise that returns `true` or `false` and not a JSON Object.

ahamdi: @ahmadmido Could you take care of this feedback

ahamdi (dismissed):

boubaker (commented):
Could you rebase the branch to be able to review it again please

ahmadmido:
> Could you rebase the branch to be able to review it again please

Done!

apps/portlet-documents/src/main/webapp/vue-app/attachment/extensions.js
```diff
@@ -54,30 +54,35 @@ export function installExtensions() {
...
-        const docPaths = activity.templateParams.DOCPATH && activity.templateParams.DOCPATH.split('|@|')
-                          || (activity.templateParams.nodePath && [activity.templateParams.nodePath]);
-        return docPaths && docPaths.length > 1;
+        extensionRegistry.registerExtension('activity', 'action', Object.assign({
+          id: 'downloadAll',
+          labelKey: 'documents.label.downloadAll',
+          isEnabled: activity => {
+            if (activity.templateParams) {
+              const docPaths = activity.templateParams.DOCPATH && activity.templateParams.DOCPATH.split('|@|')
+          || (activity.templateParams.nodePath && [activity.templateParams.nodePath]);
```
boubaker: please fix this block format

apps/portlet-documents/src/main/webapp/js/transferRulesService.js
```diff
@@ -0,0 +1,18 @@
...
+      return resp.json();
+    }
+    else {
+      throw new Error ('Error when getting transfer rules download documents status');
+    }
+  }).then(data => {
+    if (data) {
+      const downloadStatusPromise =  data.downloadDocumentStatus !== 'true';
+      return Promise.resolve(downloadStatusPromise);
+    }
```
boubaker: It should be `downloadDocumentStatus === 'true'` to be able to download, isn't it ?
**BTW**: no need to create a `new Promise`, since the value will be returned with inside a `Promise` already.

```suggestion
    return data && data.downloadDocumentStatus === 'true';
```

apps/portlet-documents/src/main/webapp/vue-app/attachment/extensions.js
ahmadmido: > please fix this block format

Done!

apps/portlet-documents/src/main/webapp/js/transferRulesService.js
ahmadmido: > It should be `downloadDocumentStatus === 'true'` to be able to download, isn't it ?
When `downloadDocumentStatus === 'true' the download action is disabled, we suspend the download.
> **BTW**: no need to create a `new Promise`, since the value will be returned with inside a `Promise` already.
Done!

boubaker (approved):

PR #1432, opened 2021-09-06
Task-48226 : Make folder icon clickable on upload documents drawer

ahmadmido:
Issue: Before this fix the folder icon on the upload documents drawer is disabled, The select folder drawer isn't displayed.
Solution: Modify the condition on displayMessageDestinationFolder method to display the select folder drawer if some of the uploaded documents (documents) aren't null.


AzmiTouil (approved):
LGTM

PR #1437, opened 2021-09-08
Task-48539 : Display info message when changing folder destination

ahmadmido:
Issue: No info message is displayed when the user selects a folder destination on the destination drawer.
Solution: The following message will be displayed after selecting a destination folder "ou attached files from ***Destination***, These files will be available for your connections once you post them"


AzmiTouil (approved):
LGTM

PR #1440, opened 2021-09-08
Task-48600: Wrong message is displayed when select files from google …

lbengayed:
…drive-Personal documents

Add label in the .properties file for the sharing file from Drive and give it the value that will be displayed.


apps/portlet-documents/src/main/webapp/vue-app/legacy-composer-attachments/components/ExoAttachments.vue
```diff
@@ -41,7 +41,7 @@
             <b v-show="fromAnotherSpaces.length > 0">
               {{ fromAnotherSpaces }}
             </b>
-            {{ $t('attachments.alert.sharing.availableFor') }} <b>{{ spaceGroupId }}</b> {{ $t('attachments.alert.sharing.members') }}
+            {{ $t('attachments.alert.sharing.filesAvailableFor') }} <b>{{ spaceGroupId }}</b> {{ $t('attachments.alert.sharing.members') }}
```
ahamdi: the task says that this key **attachments.alert.sharing.personal** has no value in the resource bundles and needed a value, why do we need to change it with another resource bundle?

lbengayed: We have two bugs in this task, once for the key 'attachments.alert.sharing.personal' i have added the value, 
The second was for the message , we need to change the message with another resource bundle, it was 'This file will be available for all members once you post it. /spaces/space_x members once you post them' but the Expected behavior was 'These files will be available for/spaces/space_x members once you post them'

ahamdi: Ok, then you need to check for the number of attached files is bigger than 1 and use the old or new label foe each case.
Also please change the current behavior and put the space display name instead of the group Id , it is more meaningful for end users.

AzmiTouil (approved):
LGTM

PR #1456, opened 2021-09-17
Task-48682: Image not displayed for a member of space after edit a no…

lbengayed:
…te page.

Simple user can't access to the image added in the note page, since the permission of the Notes folder was *:/platform/administrators. I change the permission to *:/spaceGroupId in order to the simple user can access to the image.

### DIFF
--- a/ecms-social-integration/src/main/java/org/exoplatform/ecms/uploads/HTMLUploadImageProcessorImpl.java
+++ b/ecms-social-integration/src/main/java/org/exoplatform/ecms/uploads/HTMLUploadImageProcessorImpl.java
@@ -15,6 +15,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import org.exoplatform.container.PortalContainer;
 import org.exoplatform.ecm.utils.text.Text;
+import org.exoplatform.services.cms.BasePath;
 import org.exoplatform.services.cms.impl.Utils;
 import org.exoplatform.services.cms.link.LinkManager;
 import org.exoplatform.services.jcr.RepositoryService;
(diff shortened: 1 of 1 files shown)


ecms-social-integration/src/main/java/org/exoplatform/ecms/uploads/HTMLUploadImageProcessorImpl.java
```diff
@@ -200,10 +202,11 @@ public String processSpaceImages(String content, String spaceGroupId, String ima
       Session session = sessionProvider.getSession("collaboration",
               repositoryService.getCurrentRepository());
       Node groupNode = session.getRootNode().getNode("Groups");
+      String newSpaceGroupId=spaceGroupId;
```
ahamdi: Why did you introduce this variable ?

ecms-social-integration/src/main/java/org/exoplatform/ecms/uploads/HTMLUploadImageProcessorImpl.java
```diff
@@ -224,6 +227,8 @@ public String processSpaceImages(String content, String spaceGroupId, String ima
           }
           if (imagesFolderNode.hasNode(folder)) {
             imagesFolderNode = imagesFolderNode.getNode(folder);
+            imagesFolderNode.addMixin("exo:privilegeable");
```
ahamdi: Check if the node does not have that mixin before adding it

ahamdi (dismissed):

ecms-social-integration/src/main/java/org/exoplatform/ecms/uploads/HTMLUploadImageProcessorImpl.java
lbengayed: done !

AzmiTouil (approved):

PR #1466, opened 2021-09-22
Task-48655 : empty doc / docx file attached to the post disappears

AzmiTouil:
Prior this change, when upload a file with size 0 (rare case) and then post, the file disappears and the log display errors.
We decide to limit the upload process by preventing the user from uploading a file with size 0 and by informing him by alert message.


apps/portlet-documents/src/main/resources/locale/portlet/attachments_en.properties
```diff
@@ -13,6 +13,7 @@ attachments.drawer.maxFileSize=Maximum file size : {0} MB
 attachments.drawer.maxFileCountLeft=Max files left: {0}
 attachments.drawer.maxFileCount=Max files: {0}
 attachments.drawer.maxFileSize.error=You can\u2019t upload files larger than {0} MB
+attachments.drawer.nullFileSize.error=Attached file size must be greater than 0 bytes
```
ahamdi: You can't upload an empty file
```suggestion
attachments.drawer.nullFileSize.error=You can't attach an empty file
```

ahamdi (approved):

PR #1485, opened 2021-09-30
Task-49273: Hide space navigation in mobile view and fix attachment drawer style

hassenelkamel-exo:
Prior to this change, some styles were added to attachment drawer footer. Due to those properties, the drawer in mobile view is not well displayed, so to solve this problem, we have removes those styles and we hide the spaceButtomNavigation when opening the drawer


azayati (changes requested):
verify that this changements des not impact other apps using attachment drawer

hassenelkamel-exo:
> verify that this changements des not impact other apps using attachment drawer

I was tested the other attachment drawer, no problem was detected

azayati (approved):
LGTM

PR #1493, opened 2021-10-14
Task-50193: Remove the web contributor group and publisher role from ECMS

azayati:
Prior to this change, web contributor group and publisher role are configured on ecms thus not created on Meeds, after this change since web contributor group and publisher role will be created on Meeds within social
in order to allows publisher on web contributor group to write a post on a redactional space, web contributor group and publisher role are removed from ECMS


samehaidi (approved):

rdenarie:
@azayati must this PR be closed ?

PR #1526, opened 2021-11-19
Task-43189: Shared a document - Open button behavior in Shared with me tab preview

AzmiTouil:
Prior this change, the document shared with me opens in the location of the original document instead of the shared folder


hakermi (approved):
LGTM

PR #1530, opened 2021-11-22
MAINT-48685: Disable unneeded csrf check for some read and front actions

hakermi:
Disable uneeded csrf check for some read and front actions to not end with csrf expire issue for no worthy reasons


ahamdi (changes requested):
Could you reset the formatting of those files, I can not find the exact changes in some files of this PR

hakermi:
> Could you reset the formatting of those files, I can not find the exact changes in some files of this PR

I didn't' change the formatting of these files, it's interpreted in github that way

ahamdi (approved):

PR #1540, opened 2021-12-01
MAINT-51877: Display node exo:title while it's available instead of dc:title in the files search result in the documents app

hakermi:
…
**ISSUE**: When searching a file in the documents app it shows the dc:title in the result which is not good for ux when the title has different value than the dc:title in the metadata.
**FIX**: Display exo:title instead of dc:title unless it's not availbale as a property in the target node

### DIFF
--- a/core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
+++ b/core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
@@ -1060,53 +1060,28 @@ public class Utils {
   public static String getTitleWithSymlink(Node node) throws Exception {
     String title = null;
     Node nProcessNode = node;
-    if (title == null) {
-      nProcessNode = node;
-      if (nProcessNode.hasProperty("exo:title")) {
-        title = nProcessNode.getProperty("exo:title").getValue().getString();
-      }
-      if (nProcessNode.hasNode("jcr:content")) {
-        Node content = nProcessNode.getNode("jcr:content");
-        if (content.hasProperty("dc:title")) {
-          try {
-            title = content.getProperty("dc:title").getValues()[0].getString();
-          } catch (Exception e) {
-            title = null;
-          }
-        }
-      }
-      if (title != null)
-        title = title.trim();
-    }
-    if (title != null && title.length() > 0)
-      return ContentReader.getXSSCompatibilityContent(title);
     if (isSymLink(node)) {
-      nProcessNode = getNodeSymLink(nProcessNode);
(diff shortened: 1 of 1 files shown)


core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
```diff
@@ -1067,7 +1067,7 @@ public static String getTitleWithSymlink(Node node) throws Exception {
       }
       if (nProcessNode.hasNode("jcr:content")) {
```
rdenarie: Should we move your modification in the upper if ?
So that, if there is a exo:title, we do not have to read jcr:content subnode.
It requires less compute.

rdenarie (changes requested):

core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
hakermi: Ok

core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
```diff
@@ -1065,7 +1065,7 @@ public static String getTitleWithSymlink(Node node) throws Exception {
       if (nProcessNode.hasProperty("exo:title")) {
         title = nProcessNode.getProperty("exo:title").getValue().getString();
       }
-      if (nProcessNode.hasNode("jcr:content")) {
+      if (nProcessNode.hasNode("jcr:content") && title == null) {
```
rdenarie:
```suggestion
      if (title == null && nProcessNode.hasNode("jcr:content")) {

```

core/webui/src/main/java/org/exoplatform/ecm/webui/utils/Utils.java
```diff
@@ -1088,7 +1088,7 @@ public static String getTitleWithSymlink(Node node) throws Exception {
...
```
rdenarie:
```suggestion
      if (title == null && nProcessNode.hasNode("jcr:content")) {

```

rdenarie (changes requested):
After reflexion, this function can be refactored. For example, the first if is not necessary. 
There is a lot of duplicated code.
Can we improve that ?

hakermi:
> After reflexion, this function can be refactored. For example, the first if is not necessary. There is a lot of duplicated code. Can we improve that ?

Ok

rdenarie (approved):

PR #1541, opened 2021-12-01
Task-43577: Share document: comment zone not accessible

wassimmhamdi:
Problem: The comment zone is not clickable and you can’t write. (you have to click on the text in the comment zone several times to be editable)
Fix: Add a comment by clicking on comment zone (like the case when you open the document and select share from the top menu, you can add your comment normally)


rdenarie:
Can you explain how your modification fix the problem ?
Your description is clear, but I dont understand why you remove the whole js function

rdenarie (approved):

PR #1547, opened 2021-12-06
Task-51723: When creating a new file if destination path is not set define root folder as default path.

mdiha:


SaraBoutej (approved):
LGTM

PR #1555, opened 2021-12-09
Task-52002: Analytics - Add document upload operations in…

lbengayed:
… new and old app


core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java
```diff
@@ -555,6 +569,17 @@ public synchronized Response saveAsNTFile(String workspaceName,
 
       // return uploaded file
       Document doc = getUploadedFile(workspaceName, file, mimetype);
+      String eventName = null;
```
mdiha:
```suggestion
      String eventName = source.equals(OLD_APP) ? UPLOAD_DOC_OLD_APP : UPLOAD_DOC_NEW_APP;
```

mdiha (approved):

PR #1587, opened 2022-01-03
MAINT-52375: Avoid sql/injection in document service

hakermi:
ISSUE: When calling rest service /portal/rest/documents/folder?folder=value, The user can forge his own query to acces other documents. even if jcr sql is readonly, and that the returned data is checked with ACLs, there is no risk for data modification or exposure. but it is better to clean the variable folder before using it.
FIX: This PR should escape the injected string to prevent any dangerous chars that can cheat the sql interpreter.


rdenarie (changes requested):
I think we dont need a new class for that. Put the function in DocumentService class
If and when another class need that code, we can extract

rdenarie (approved):

PR #1588, opened 2022-01-03
Task-52369: cannot upload a new version of a document when its name includes the character "+".

montassarhamadi:
Problem: Before this fix, it was impossible to upload a new version of the document in file explorer if the filename contains a +. when checking the existence of the file, the field currentFolder contains the filename, and the + is read as a space.

Fix: This change use function encodeURIComponent for variable current folder, so that, the + is transformed to %2B and then in the rest service, we decode it with URLDecoder.decode, and we replace the parameter string "UTF-8" by StandardCharsets.UTF_8 to avoid errors in code smells, so that we are able to see the +.


apps/portlet-explorer/src/main/webapp/javascript/eXo/ecm/MultiUpload.js
```diff
@@ -283,7 +283,7 @@
 	function cleanName(oldName) {
 	
 		if (!oldName || oldName==null ||oldName==undefined) return oldName;
-		specialChar = "[]/'\":;";
+		specialChar = "[]'\":;";
```
rdenarie: As explained yesterday, you cannot remove a special character like this. It is a major risk to create side effects

rdenarie (changes requested):

rdenarie (changes requested):
In addition, can you fix the sonar code smell ?
You can use this to replace the string UTF-8 : StandardCharsets.UTF_8

apps/portlet-explorer/src/main/webapp/javascript/eXo/ecm/MultiUpload.js
montassarhamadi: ok, Romain .I left it as is and used the function encodeURIComponent to encode the special characters in URI which be affected in param "currentFolder".

rdenarie (approved):

rdenarie (changes requested):
Code is ok
But, can you update your PR description ? 

The first line should describe what is the problem, but you should add more context ?
Your first line should start by something like "Before this fix, it was impossible to upload a new version of the document in file explorer if the filename contains a +. when checking the existence of the file, the field currentFolder contains the filename, and the + is read as a space"

The second line should describe how you fix the problem : 
"This change use function encodeURIComponent for variable current folder, so that, the + is transformed to %2B and then in the rest service, we decode it with URLDecoder.decode, so that we are able to see the +"

rdenarie (approved):

PR #1600, opened 2022-01-07
Task-52482: Add new attirbute { Creation Type } in data of fileCreate…

lbengayed:
…d (#1586)

* Task-52482: Add operation fileCreated analytics for upload document with new attirbute { Creation Type }


mkrout (approved):
LGTM

PR #1613, opened 2022-01-13
MAINT-43884: Fix new_folder i18n value in file attachment drawer

hakermi:
**ISSUE**: The `getI18nTitle` function wasn't take in consideration the title with underscore when building the key label in order to offer a translation for the title.
**FIX**: Update the used regex of label key building to support titles with underscore to allow translate the new_folder label

### DIFF
--- a/apps/portlet-documents/src/main/resources/locale/portlet/documents_fr.properties
+++ b/apps/portlet-documents/src/main/resources/locale/portlet/documents_fr.properties
@@ -60,8 +60,8 @@ Folder.label.sites=Sites
 Folder.label.shared=Shared
 Folder.label.newfolder=Nouveau Dossier
 
-Category.label.MyDrives=My Drive
-Category.label.MySpaces=Mes Espaces
+Category.label.MyDrives=Mes lecteurs
+Category.label.MySpaces=Mes espaces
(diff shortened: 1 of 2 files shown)


apps/portlet-documents/src/main/resources/locale/portlet/documents_fr.properties
```diff
@@ -44,6 +44,8 @@ Drives.label.DataProtection=Protection des donn\u00E9es
 Drives.label.Guests=Invit\u00E9s
 Drives.label.Management=Gestion
 
+Folder.label.newfolder=Nouveau Dossier
```
rdenarie:
```suggestion
Folder.label.newfolder=Nouveau dossier
```

What about the english translation ?

rdenarie (changes requested):

apps/portlet-documents/src/main/resources/locale/portlet/documents_fr.properties
hakermi: already added in a previous PR but there was still no crowdin sync but no the sync is done

rdenarie (approved):

PR #1616, opened 2022-01-17
Task-49268: text area components are no longer extensible

wassimmhamdi:
Problem: in CMS setting and backoffice, text areas in web contents and list templates aren't extensible in term of hight
Fix: I set the two proprieties of Textarea which are width to 95% and max-height to unset in web contents and list templates to be extensible in term of hight


rdenarie (changes requested):
Can you update your description ? The second part does not explain how you fix the problem. It explain what you fix.

apps/resources-wcm/src/main/webapp/skin/less/ecms/portlets/explorer/ecms-explorer.less
```diff
@@ -714,8 +714,9 @@
 			textarea {
 				font-size: 13px;
 				height: 175px;
-				width: 90%;
+				width: 95%;
```
rdenarie: Should we keep 90% ? as there is not problem on the witdh, i think we should keep this unmodified
So for consistency, we should use also 90% in the other less file

rdenarie (changes requested):

apps/resources-wcm/src/main/webapp/skin/less/ecms/portlets/administration/ecms-administration.less
```diff
@@ -43,7 +43,9 @@
 				.controls{
 					.textarea{
 						height: 150px;
+						width: 95%;
```
rdenarie: As said in my last comment, should we keep 90% here to make it consistent ?

rdenarie (changes requested):

rdenarie (approved):

PR #1621, opened 2022-01-19
MAINT-52789: Reindex nodes on jcr node_moved ExtendedEvent to allow searching moved files in different paths after clone or upload inside documents app

hakermi:
…
**ISSUE**: When clone files from Google drive using gdriveimporter addon and move files into different folders and sub-folders the indexing engine doesn't re-index the files again which caused a wrong search result and
the disappearance of these files from the search result.
**FIX**: Add a new `ActionConfig` for the `exo:move` and the node_moved event and add the `eventtype` in `FileIndexerAction` to listen the event that dispatched from from the `SessionActionInterceptor`


ecm-wcm-extension/src/main/webapp/WEB-INF/conf/dms-extension/dms/dms-actions-configuration.xml
```diff
@@ -67,8 +67,23 @@
...
+                      <field  name="srcWorkspace"><string>collaboration</string></field>
+                      <field  name="srcPath"><string>/</string></field>
+                      <field  name="isDeep"><boolean>true</boolean></field>
+                      <field  name="lifecyclePhase">
+                        <collection type="java.util.ArrayList">
+                          <value><string>node_moved</string></value>
+                        </collection>
+                      </field>
+                    </object>
+                  </value>
```
rdenarie, line 86: Can you explain what this action is for ? I'm not sure to understand.
Thanks

rdenarie (changes requested):

ecm-wcm-extension/src/main/webapp/WEB-INF/conf/dms-extension/dms/dms-actions-configuration.xml
hakermi: This action need to be registered in `ActionCatalog` : https://github.com/exoplatform/jcr/blob/develop/exo.jcr.component.core/src/main/java/org/exoplatform/services/jcr/impl/ext/action/SessionActionInterceptor.java#L56 
declared for exo:move type and dispatched event is node_moved its name is `moveNode` which i added it in eventTypes of the `FileIndexerAction`
https://github.com/exoplatform/jcr/blob/develop/exo.jcr.component.core/src/main/java/org/exoplatform/services/jcr/impl/ext/action/SessionActionInterceptor.java#L512  

if we don't register it the event will not be launched :
https://github.com/exoplatform/jcr/blob/develop/exo.jcr.component.core/src/main/java/org/exoplatform/services/jcr/impl/ext/action/SessionActionInterceptor.java#L563

rdenarie (approved):

PR #1622, opened 2022-01-20
MAINT-43884: Add an i18n translated title when adding a new folder

hakermi:
ISSUE: when adding a new folder an input field is appeared shows taking a default new_folder value as a name which is not a good us even if the label will be translaed after saving the folder
FIX: Add a translated title for the new_folder since the pre_saving of the new folder

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
@@ -827,7 +827,7 @@ export default {
     addNewFolder() {
       if (!this.creatingNewFolder) {
         this.creatingNewFolder = true;
-        this.newFolderName = 'new_folder';
+        this.newFolderName = this.$t('Folder.label.newfolder');
         this.folders.unshift({
           id: 'new_folder',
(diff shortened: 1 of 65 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
```diff
@@ -827,7 +827,7 @@ export default {
     addNewFolder() {
       if (!this.creatingNewFolder) {
         this.creatingNewFolder = true;
-        this.newFolderName = 'new_folder';
+        this.newFolderName = this.$t('Folder.label.newfolder');
```
rdenarie, line 837: should it be "nouveau_dossier" in french ? should we keep the default name without space ?
How the folder is named if I save "Nouveau dossier" ?

rdenarie (approved):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
hakermi: it will be saved **nouveau dossier** in french with space and it's okay with jcr i tested it even with Cyrillic languages

PR #1629, opened 2022-01-25
MAINT-52789: Reindex nodes on jcr node_moved ExtendedEvent to allow searching moved files in different paths after clone or upload inside documents app

hakermi:
…
ISSUE: When clone files from Google drive using gdriveimporter addon or upload inside docs app and move files into different folders and sub-folders the indexing engine doesn't reindex the files again which caused a wrong search result and
the disappearance of these files from the search result.
FIX: Add a new `ActionConfig` for the `exo:move` and the node_moved event and add the eventtype in `FileIndexerAction` to listen the event that dispatched from the `SessionActionInterceptor`

### DIFF
--- a/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-search-configuration.xml
+++ b/core/core-configuration/src/main/webapp/WEB-INF/conf/wcm-core/core-search-configuration.xml
@@ -129,7 +129,7 @@
                                 <value>
                                     <object type="org.exoplatform.services.jcr.impl.ext.action.ActionConfiguration">
                                         <field  name="workspace"><string>collaboration</string></field>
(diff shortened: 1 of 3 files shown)


eromdhaneexo (dismissed):

rdenarie (changes requested):
@hakermi in your description, you mention exo gdrive addon.
Does the problem is visible without the addon ? Is it an ecms problem or gdrive addon problem ?

hakermi:
> @hakermi in your description, you mention exo gdrive addon. Does the problem is visible without the addon ? Is it an ecms problem or gdrive addon problem ?

The problem in the ecms not in the addon, but it was reported in context of using this addon i will update the description.

rdenarie (approved):

PR #1631, opened 2022-01-25
MAINT-43884: Fix new_folder i18n value in file attachment drawer

hakermi:
* MAINT-43884: Fix new_folder i18n value in file attachment drawer
ISSUE: The getI18nTitle wasn't take in consideration the title with underscore when building the key label in order to offer a translation for the title.
FIX: Update the used regex of label key building to support titles with underscore to allow translate the new_folder label

* MAINT-43884: Add an i18n translated title when adding a new folder
ISSUE: when adding a new folder an input field is appeared shows taking a default new_folder value as a name which is not a good us even if the label will be translaed after saving the folder
FIX: Add a translated title for the new_folder since the pre_saving of the new folder


rdenarie (changes requested):
Can you update the PR description ? there is 3 times the message

hakermi:
> Can you update the PR description ? there is 3 times the message

done

rdenarie (approved):

PR #1651, opened 2022-02-02
TASK-53813: Add support (view and edit) for docxf and oform file types with onlyoffice

hakermi:
Prior to this change, docxf and oform types are not supported with our onlyoffice integration. This PR should make the onlyoffice able to view and edit these files and also adding these types to the `OfficeThumbnailPlugin` to allow creating thumbnails.


mkrout (approved):
LGTM

PR #1656, opened 2022-02-11
Task 54048: Share document from a user having a read only permission.

montassarhamadi:
Problem: before this fix, When the members of space have the permission read-only in a document, the icon of a share document was always showing. So we need to add this icon just when the members have edit permission.
    
Fix: in this fix, we add the filter CanSetPropertyFilter to ShareDocumentsComponent for be added to toolbarAction  just when the user or group has edit permission.


ecms-social-integration/src/main/java/org/exoplatform/ecm/webui/component/explorer/rightclick/manager/ShareDocumentsComponent.java
```diff
@@ -20,6 +20,8 @@
 import java.util.List;
 
 import org.exoplatform.ecm.webui.component.explorer.UIJCRExplorer;
+import org.exoplatform.ecm.webui.component.explorer.control.filter.CanEditDocFilter;
```
rdenarie:
```suggestion
```

rdenarie (changes requested):

rdenarie (approved):

PR #1659, opened 2022-02-18
Task-51341: many emails received after editing a watched document.

montassarhamadi:
Problem: Before this fix, when editing a watched file from users. The watcher was receiving many emails indicating that there is changes on the file.SO this problem occurred because the process of observation was on the node and not the child node which is the jcr:content, therefore, any action that happens in the file, can change many properties even if we open and close the file without changing it. So many events fired and On every event, there's an email.
Fix: In this fix, we changed the process of the observation on the content of the file and we let the list of watchers in property EMAIL_WATCHERS_PROP of the watched file. So  any changes in the content of the watched file will fire an event which be received by the EmailNotifyCloudDocumentListener and finally this last will send  one email to each watcher if exist .

### DIFF
--- a/core/webui-clouddrives/src/main/java/org/exoplatform/services/cms/clouddrives/webui/watch/EmailNotifyCloudDocumentListener.java
+++ b/core/webui-clouddrives/src/main/java/org/exoplatform/services/cms/clouddrives/webui/watch/EmailNotifyCloudDocumentListener.java
@@ -57,6 +57,7 @@ import org.exoplatform.services.security.Identity;
 import org.exoplatform.services.security.IdentityRegistry;
 import org.exoplatform.services.security.MembershipEntry;
 import org.exoplatform.services.wcm.core.NodeLocation;
+import org.exoplatform.services.wcm.core.NodetypeConstant;
 import org.exoplatform.services.wcm.utils.WCMCoreUtils;
 import org.exoplatform.web.url.navigation.NodeURL;
 import org.exoplatform.webui.application.WebuiRequestContext;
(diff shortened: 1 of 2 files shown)


core/webui-clouddrives/src/main/java/org/exoplatform/services/cms/clouddrives/webui/watch/WatchCloudDocumentServiceImpl.java
```diff
@@ -154,7 +156,7 @@ public void watchDocument(Node documentNode, String userName, int notifyType) th
       }
     }
 
-    if (listener != null && documentNode.isNodeType(JCRLocalCloudDrive.ECD_CLOUDFILE)
+    if (listener != null && documentNode.isNodeType(NodetypeConstant.NT_FILE)
```
rdenarie: Can you explain why you changed this test ?

core/webui-clouddrives/src/main/java/org/exoplatform/services/cms/clouddrives/webui/watch/EmailNotifyCloudDocumentListener.java
```diff
@@ -156,7 +159,7 @@ private Message createMessage(String receiver, MessageConfig messageConfig) thro
     String documentLink;
     if (WebuiRequestContext.getCurrentInstance() != null) {
       documentLink = getViewableLink();
-    } else if (node.isNodeType(JCRLocalCloudDrive.ECD_CLOUDFILE)
+    } else if (node.isNodeType(NodetypeConstant.NT_FILE)
```
rdenarie: Can you explain why you changed this test ?

rdenarie (changes requested):

rdenarie (changes requested):
Before the fix, if createMessage or sendMessage throw an error, the for execution continue, and the other users of the list are notified
After the fix, if createMessage or sendMessage throw an error, the for execution stops, and the other users of the list are not notified.

Please modify the behaviour, so that the try in inside the for

core/webui-clouddrives/src/main/java/org/exoplatform/services/cms/clouddrives/webui/watch/EmailNotifyCloudDocumentListener.java
```diff
@@ -113,18 +114,26 @@ public void onEvent(EventIterator arg0) {
...
+        } catch (Exception e) {
+            if (LOG.isErrorEnabled()) {
+              LOG.error("Unexpected error", e);
+          }
+        }
+      }
+    } catch (RepositoryException e) {
         if (LOG.isErrorEnabled()) {
-          LOG.error("Unexpected error", e);
+          LOG.error("jcr error occured", e);
```
rdenarie:
```suggestion
          LOG.error("Unable to get node location", e);
```

rdenarie (changes requested):

rdenarie (approved):

rdenarie (changes requested):
Can you fix the sonar issue ?

rdenarie (changes requested):
What if NodeLocation.getNodeByLocation generate an error ?

Your previous solution was better. You just had to extract 
`try {
        Message message = createMessage(receiver, messageConfig);
        mailService.sendMessage(message);
      } catch (Exception e) {`
in a submethod
Can you rollback your last modification ?

montassarhamadi:
> 

What do you suggest as a name for the submethod??

rdenarie (approved):

PR #1663, opened 2022-02-21
TASK-53835 Fix Analytics Profile for JCR Action configuration

boubaker:
Prior to this this change, the Analytics Listener configuration is added even when analytics addon isn't installed. This modification ensures that this listener isn't added in such case, by introducing profiles=analytics value in configuration file.


AzmiTouil (approved):
LGTM

PR #1674, opened 2022-03-08
Task-54742: Avoid display files attachments when the activity type is not simple, file or link activity

abenmbarek:
Before this change, for any type of posted activity, when files attachments are uploaded, they will be displayed in the created activity body.
After this change, we will avoid displaying the files attachments when the activity type is not simple, file or link activity.


azayati (approved):
LGTM after QA validation

PR #1679, opened 2022-03-14
Task-55243: Keep attached files when the activity type is not simple, file or link activity

azayati:
Prior to this change, attached files are not kept the activity type is not simple, file or link activity. This PR will revert commit 51264950a5f2f373cb0f85ec5a02a18f9493796c in order to keep attached files in that case.


abenmbarek (approved):
LGTM after QA validation

PR #1686, opened 2022-03-17
Task-55028: Notification of call recording or recording failure

montassarhamadi:
improvement: we added an optional request parameter to the parameters of the redirection method for the following reason: when we add this optional parameter, the redirection is done to the url without param userId in the case of a user-to-user call record.

### DIFF
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <artifactId>ecms</artifactId>
     <groupId>org.exoplatform.ecms</groupId>
-    <version>6.3.x-SNAPSHOT</version>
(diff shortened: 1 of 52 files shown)


rdenarie:
@mkrout Can you have a check on this modification ? 
@Montassar-Hamadi I'm not sure we should change something here. 
Can you explain to Krout what you have in mind ?

montassarhamadi:
> @mkrout Can you have a check on this modification ? @Montassar-Hamadi I'm not sure we should change something here. Can you explain to Krout what you have in mind ?

No need to merge this PR @rdenarie in module ecms because I created it when I suggested to you the second solution for this task. But now the implementation is changed and I will close this now

PR #1698, opened 2022-03-31
Task-54703: Document: link in the email notification is not functional

hchaabene:
Prior to this change, when post a document in group space , then select the document and click in Watch then check Notify by email ,then click to document and modify it.
This is due to the fact that mail is send but the link sometimes the link is empty and sometimes it not redirected at the true destination.
After this change, we ensure that email send a link to notify users that document is changed and that link to redirect in the destination document.


rdenarie (changes requested):
Does getDrive is still used ? if not perhaps we could remove it.
Can you show an example of link generated with your fix ?

rdenarie (approved):

PR #1708, opened 2022-04-05
TASK-55665 Define Administration Menu Add-on Categories and URIs

boubaker:
This modification will use the new Administration Menu configuration extensibility to define menu item entries.


AzmiTouil (approved):
LGTM

PR #1723, opened 2022-04-21
TASK-55630 Fix Copying enrolled node in publication lifecycle

boubaker:
Prior to this change, when copy/paste a content that is enrolled in Publication lifecycle, the property 'publication:liveRevision' (of type REFERENCE) references the original content published version. Consequently when attempting to delete original content from JCR, a ConstraintViolationException is throw which will prevent content deletion. This change will enroll again the copied content inside the publication lifecycle to cleanup references to original node and will make the copied content as Draft, in contrary to the previous result which was leading to have a copy of content in 'Published' state.

(**Remaining**: for previously copied contents and placed in Trash, we will have to add a specific workaround to allow deleting previously referenced contents )


boubaker:
Wrong TASK identifier, replaced by https://github.com/exoplatform/ecms/pull/1724

PR #1746, opened 2022-05-12
TASK-58517: Remove related dlp base code

azayati:
Prior to this change, dlp dependancy is added into ecms and some dlp connectors implemented inside ecms addon,
This PR should remove all related dlp base code and move it to the dlp addon


core/services/src/main/java/org/exoplatform/services/attachments/utils/Utils.java
```diff
@@ -20,7 +20,7 @@
 
 public class Utils {
 
-  private static final String DLP_QUARANTINE_FOLDER = "Quarantine";
+  public static final String DLP_QUARANTINE_FOLDER = "Quarantine";
```
ahamdi: We need to remove this reference too

ecms-social-integration/src/main/java/org/exoplatform/ecms/activity/processor/ActivityAttachmentProcessor.java
```diff
@@ -124,7 +124,7 @@ private String getTitle(Node contentNode) throws RepositoryException {
   }
 
   private boolean isQuarantinedItem(Node node) throws RepositoryException {
-    return node.getPath().startsWith("/" + FileDlpConnector.DLP_QUARANTINE_FOLDER + "/");
+    return node.getPath().startsWith("/" + Utils.DLP_QUARANTINE_FOLDER + "/");
```
ahamdi: We need to remove this reference too

ahamdi (changes requested):

ecms-social-integration/src/main/java/org/exoplatform/ecms/activity/processor/ActivityAttachmentProcessor.java
hakermi: As discussed  in we keep this check, and only rename the variable.

ahamdi (approved):

PR #1749, opened 2022-05-13
TASK-57198: Fix jcr session already closed while saving news article

hakermi:
Prior to this change, both `HtmlUploadImageProcessor` and other services such as news using `sessionProviderService` to get the session provider. Since the `sessionProviderService` shares the `sessionProivder` in a local thread, closing the session provider by the processor made it inaccessible and not useful for other services in the same thread.
This PR removes the `sessionprovider` close in the `HTMLUploadImageProcessor` to avoid such exception.


azayati (approved):
LGTM

PR #1757, opened 2022-05-16
Task-56988 : Update processUpload rest API to create folder when it doesn't exist

AzmiTouil:
Prior this change, we cannot upload a file to a folder that does not exist
Update : Update processUpload rest API to create folder when it doesn't exist


core/connector/src/main/java/org/exoplatform/ecm/connector/platform/ManageDocumentService.java
```diff
@@ -351,10 +326,10 @@ public Response createFolder(@QueryParam("driveName") String driveName,
 
       Node newNode = node.addNode(name, folderNodeType);
 
-      if (!newNode.hasProperty("exo:title")) {
-        newNode.addMixin("exo:rss-enable");
+      if (!newNode.hasProperty(NodetypeConstant.EXO_TITLE)) {
```
boubaker:
```suggestion
      if (!newNode.hasProperty(NodetypeConstant.EXO_TITLE) && newNode.canAddMixin(NodetypeConstant.EXO_RSS_ENABLE)) {
```

apps/portlet-documents/src/main/webapp/vue-app/attachment/extensions.js
```diff
@@ -40,6 +41,13 @@ export function installExtensions() {
     rank: 10,
   });
 
+  extensionRegistry.registerExtension('CommonCkeditor', 'ckeditor-extensions', {
+    id: 'selectImage',
+    extraPlugin: 'selectImage',
+    extraToolbarItem: 'selectImage',
+    rank: 20,
+  });
```
boubaker:
```suggestion
  extensionRegistry.registerExtension('WYSIWYGPlugins', 'image', {
    id: 'selectImage',
    extraPlugin: 'selectImage',
    extraToolbarItem: 'selectImage',
    rank: 20,
  });
```

IMO, this extension name isn't adequate, you should add a specific extension name to image like `CkeditorPlugins` /  `image`

core/connector/src/main/java/org/exoplatform/ecm/connector/platform/ManageDocumentService.java
```diff
@@ -686,13 +661,22 @@ private Node getNode(String driveName, String workspaceName, String currentFolde
...
         node = node.getNode(folder);
-        if (node.isNodeType(NodetypeConstant.EXO_SYMLINK)) node = linkManager.getTarget(node);
-      } else if (node.isNodeType(NodetypeConstant.EXO_SYMLINK)) {
-        node = linkManager.getTarget(node).getNode(folder);
       } else {
-        return node;
+        // create new folder
+        String name = Text.escapeIllegalJcrChars(org.exoplatform.services.cms.impl.Utils.cleanString(folder));
+        name = (StringUtils.isEmpty(name)) ? DEFAULT_NAME : name;
+        Node newNode = node.addNode(name, NodetypeConstant.NT_FOLDER);
```
boubaker, line 670: which session is used to create folder, what are the default permissions added into this folder ? Is it 'any' added by default here or does it inherit default permissions from parent. Does a dedicated permission is added to the current user ? IMO it shouldn't, because the user can revoke its membership in space and consequently, he shouldn't be able to access this folder neither modify its content.

apps/portlet-documents/src/main/resources/locale/portal/webui_nl.properties
```diff
@@ -0,0 +1,58 @@
+#
+# Copyright (C) 2014 eXo Platform SAS.
```
boubaker: please update the License Headers : 
1/ AGPL instead of LGPL
2/ 2022 instead of 2014

core/connector/src/main/java/org/exoplatform/ecm/connector/platform/ManageDocumentService.java
AzmiTouil: checked , the folder inherit default permissions from parent, The permissions set for the admins and the space members

boubaker (approved):

PR #1758, opened 2022-05-17
Task-57296: Infinite loading of the folder when changing the location  of a file .

montassarhamadi:
Problem: Infinite loading of the folder when changing the location of a file.
Fix : assigned to the variable driveExplorerInitializing value false to stop loading when the initialization of Destination folders is finished.

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
@@ -575,10 +575,10 @@ export default {
           }
           //if both default drive and default folder exist
           if (defaultFolder) {
-            this.openFolder(defaultFolder).then(() => {
-              this.$root.$emit('attachments-default-folder-path-initialized', this.getRelativePath(self.selectedFolderPath), this.schemaFolder);
-              this.driveExplorerInitializing = false;
-            });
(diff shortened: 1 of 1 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
```diff
@@ -600,6 +600,7 @@ export default {
             
           } else {
             this.$root.$emit('attachments-default-folder-path-initialized', '/', this.currentDrive.title);
+            this.driveExplorerInitializing = false;
```
boubaker, line 603: We may need to change as well the code in https://github.com/exoplatform/ecms/blob/7b256459ed31f52a10c3b3cedc098ddee34cea94/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue#L578-L581

to make it as follow : 
```javascript
                this.openFolder(defaultFolder)
                  .then(() => {
                    this.$root.$emit('attachments-default-folder-path-initialized', this.getRelativePath(self.selectedFolderPath), this.schemaFolder);
                  }).finally(() => this.driveExplorerInitializing = false);
```

The `catch` case isn't handled by original code, thus either when the Promise gets resolved or rejected, in all cases, we need to stop loading UI effect.

montassarhamadi:
done and tested @boubaker

boubaker (approved):

PR #1762, opened 2022-05-17
TASK-57240: Fix refreshing uploaded files when editing an Activity

Amine0008:
ISSUE: Sometimes when attaching newly created doc files, the attachments section in the "Add Documents" drawer isn't updated. 
FIX: When creating a document in the "Add Documents" drawer, we will not push the created doc to the uploadedFiles array since the push will emit the "entity-attachments-updated" event which reinitializes the attachments list.


boubaker (approved):

azayati (changes requested):
The title and description should be cleaned IMO

azayati (approved):
LGTM

PR #1764, opened 2022-05-17
TASK-55861: Add Content Length HTTP Header when reading Image Thumbnail

akhanfir:
Prior to this change, the image thumbnail isn't completely displayed to end user due to missing Content-Length HTTP Header (reproduced on Windows OS only). This fix will ensure to add the missing HTTP Header to ensure browser compatibility with image content display.


hassenelkamel-exo (approved):
LGTM

PR #1766, opened 2022-05-18
Task-57195: cleaning node name that contains "."

montassarhamadi:
Issue: an exception is thrown when creating a node which has the character "." in name.
Fix:  Replace `.` character with `-` in reusable method `Utils.cleanName`


boubaker (commented):
Need : 
1/ A unit Test
2/ Redefine title ( it must not re-use the same title as Task but the description of the fix as suggested here https://developer.exoplatform.org/#id-commits-messages-best-practices )
3/ Review description please

boubaker (approved):
Please user new title & description when merging. You shouldn't reference other addons names in commits.

PR #1769, opened 2022-05-18
Task-57296 Progress bar still displayed when changing the location of a file

montassarhamadi:
Problem: The progress bar still displayed when changing the location of a file.
Fix : stop displaying the progress bar when the location change is made successfully


boubaker (approved):
Please use new title & description while committing

PR #1774, opened 2022-05-19
Task-56752: Bad display for cloud storage page on Firefox

hchaabene:
Prior to this change, when go to left Navigation menu -> administration-> documents -> cloud storage then hover in table , a scroll bar is displayed below the table , then page title is not well displayed.
Fix: i remove overflow hidden in the class "Title" , then remove class table in the selector v-simple-table 
After this change, that title is displayed clearly and a scroll bar is not be displayed.


boubaker (approved):

boubaker:
please add PR description

PR #1787, opened 2022-05-24
TASK-57175: Add getDocumentsTransferRules function in transferRulesService.js

hakermi:
Prior to this change, no backend restriction made for download documents depends on tranfer rules.
This PR should add a restriction depends on transfer rules choose settings

### DIFF
--- a/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
+++ b/apps/portlet-documents/src/main/webapp/js/transferRulesService.js
@@ -15,3 +15,16 @@ export function getTransfertRulesDownloadDocumentStatus() {
     }
   });
 }
+
+export function getDocumentsTransferRules() {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/transferRules/getTransfertRulesDocumentStatus`, {
+    credentials: 'include',
+    method: 'GET',
+  }).then(resp => {
+    if (!resp || !resp.ok) {
+      throw new Error('Response code indicates a server error');
+    } else {
+      return resp.json();
+    }
+  });
+}


core/services/src/main/java/org/exoplatform/services/cms/webdav/WebDavServiceImpl.java
```diff
@@ -294,6 +300,13 @@ public Response get(@PathParam("repoName") String repoName,
       }
       return Response.serverError().build();
     }
+    SettingValue<?> downloadDocumentSettingValue =
+                                                 settingService.get(org.exoplatform.commons.api.settings.data.Context.GLOBAL.id("downloadDocumentStatus"),
+                                                                    Scope.APPLICATION.id("downloadDocumentStatus"),
+                                                                    "exo:downloadDocumentStatus");
```
boubaker: It would be better to not use the implementation details of Transfer Rule API here, but define a common API JAR where to define the Service API layer in order to reuse it inside other implementation JARs

hakermi: @boubaker where i can define the common api of transferRules ?  in ecms or commons module or gatein-portal beside the settingservice ?

boubaker: ok, I see the problem, there is no Service layer for Transfer Rules API, in addition, the feature is mean't to change options in UI only without restrictions in backend. Thus, let's fix the frontend only.
( Bad architecture design already made, it has to be prioritized in a dedicated Feature Request )

hakermi: done

boubaker (approved):

PR #1791, opened 2022-05-25
TASK-57545: fix file upload using safari

Amine0008:
ISSUE: IOS devices use ISO-8859-1 to encode character which intoduces problems when uploading a file containing french characters (different code for non standard alphabet characters )
FIX:  used already predefined method to clean the name

### DIFF
--- a/apps/portlet-documents/package.json
+++ b/apps/portlet-documents/package.json
@@ -29,7 +29,6 @@
   "dependencies": {
     "anchorme": "^1.1.2",
     "autosize": "^4.0.2",
-    "axios": "^0.21.2",
-    "utf8": "^3.0.0"
+    "axios": "^0.21.2"
   }
 }
--- a/apps/portlet-documents/package.json
+++ b/apps/portlet-documents/package.json
@@ -29,6 +29,7 @@
   "dependencies": {
     "anchorme": "^1.1.2",
     "autosize": "^4.0.2",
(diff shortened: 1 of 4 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/main.js
```diff
@@ -14,7 +14,7 @@ const urls = [
   `/portal/rest/i18n/bundle/locale.portlet.attachments-${lang}.json`,
   `/portal/rest/i18n/bundle/locale.portlet.documents-${lang}.json`
 ];
-
+eXo.env.portal.utf8 = require('utf8');
```
boubaker: no third party library is allowed to be imported using npm packages. You should find a solution with available APIs

apps/portlet-documents/package.json
```diff
@@ -29,7 +29,6 @@
   "dependencies": {
     "anchorme": "^1.1.2",
     "autosize": "^4.0.2",
-    "axios": "^0.21.2",
-    "utf8": "^3.0.0"
```
boubaker: to revert this change please

Amine0008: all changes to package.json has been reverted to its  original state in the feature/maintenance branch

boubaker (approved):

PR #1808, opened 2022-06-23
TASK-58025: Remove MalwareDetectionJcr Connector 

hakermi:
Prior to this change, MalwareDetectionJcr connector was created inside the ecms module.
This PR should remove it from ecms and add it inside the anti-malware addon


azayati (approved):
LGTM

PR #1812, opened 2022-06-30
Task-58506: Add Connect your drive drawer

abenmbarek:
After this PR, we will be able to display a drawer containing the list of drive connectors without any behavior in this task.


azayati (approved):
LGTM

PR #1813, opened 2022-06-30
Task-58643: Related the attached file thumbnails to attachment file extension

sofyenne:
ISSUE : when user create process and attached a file all attached file thumbnails displayed as media , the problem is when clean name of uploaded document   can't extract the extension from the name .
FIX : modified clean name method to correctly clean the name file

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
@@ -701,15 +701,22 @@ public class Utils {
 
   public static String cleanName(String oldName) {
     if (StringUtils.isEmpty(oldName)) return oldName;
-    String specialChar = "&#*@.'\"\t\r\n$\\><:;[]/%|";
+    String specialChar = "&#*@'\"\t\r\n$\\><:;[]/%|";
     StringBuilder ret = new StringBuilder();
     for (int i = 0; i < oldName.length(); i++) {
       char currentChar = oldName.charAt(i);
       if (specialChar.indexOf(currentChar) > -1) {
-        ret.append('-');
+        ret.append('_');
       } else {
         ret.append(currentChar);
       }
+    }
+    if(ret.indexOf(".") > -1 && ret.indexOf(".") != ret.lastIndexOf(".") ){
+      String extention = ret.substring(ret.lastIndexOf("."));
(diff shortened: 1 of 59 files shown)


core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
```diff
@@ -701,16 +701,23 @@ public static String cleanString(String str) {
...
     for (int i = 0; i < oldName.length(); i++) {
       char currentChar = oldName.charAt(i);
       if (specialChar.indexOf(currentChar) > -1) {
-        ret.append('-');
+        ret.append('_');
       } else {
         ret.append(currentChar);
       }
     }
+    if(ret.indexOf(".") > -1 && ret.indexOf(".") != ret.lastIndexOf(".") ){
```
ahamdi: You can just substring the extension in the beginning of the cleanup process and then do the cleanup includin hg the dot (.) character.

core/services/src/test/java/org/exoplatform/services/cms/impl/UtilsTest.java
```diff
@@ -48,27 +49,29 @@ public void testCleanName() {
...
-    assertEquals(titleClean12,"test-test");
-    assertEquals(titleClean13,"test-test");
-    assertEquals(titleClean14,"test-test");
-    assertEquals(titleClean15,"test-test");
-    assertEquals(titleClean16,"test-test");
-    assertEquals(titleClean17,"test-test");
-    assertEquals(titleClean18,"test-test");
-    assertEquals(titleClean19,"test-test");
-    assertEquals(titleClean20,"test-test");
+    assertEquals(titleClean1,"test _ test");
```
ahamdi: why there are spaces before/after filename ?

ahamdi (changes requested):

core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
sofyenne: Done !

core/services/src/test/java/org/exoplatform/services/cms/impl/UtilsTest.java
sofyenne: done ! space removed !

ahamdi (approved):

PR #1814, opened 2022-07-01
Task-58505: Implement cloud drive settings portlet in user settings

abenmbarek:
With this change, we will implement cloud drive settings portlet and add it as a block  in user settings which will allow users to connect to their could drives using different connectors.


azayati (dismissed):
LGTM

azayati (approved):
LGTM

PR #1815, opened 2022-07-04
Task-58506: Implement cloud drive settings drawer

abenmbarek:
With this change, we will implement cloud drive settings drawer which will display a drawer containing the list of fake one drive, and google drive connectors.


azayati (approved):
LGTM

PR #1820, opened 2022-07-13
TASK-50471: Add createCustomThumbnail function in thumbnail service

hakermi:
Prior to this change, the existing scale function in thumbnail service are linked to jcr storage, each created thumbnail will be stored in jcr.
THis PR should add a new createCustomThumbnail function to scale images to custom given dimension without store in jcr


mkrout (approved):
LGTM

PR #1825, opened 2022-07-19
Task-58742: Display dynamically onedrive and gdrive connectors

abenmbarek:
Prior to this change, we display statically onedrive and gdrive connectors directly from ECMS codebase. After this commit we ensure that :

- ECMS codebase implements only the cloud drive connectors framework with the cloud drive setting drawer portlet
- Cloud-drive-connectors addon codebase implements onedrive and gdrive connectors and injects them into the cloud drive setting drawer portlet.


azayati (dismissed):
LGTM

azayati (approved):
LGTM

PR #1830, opened 2022-07-21
Task-58507: Add connection status after cloud drive connect action

abenmbarek:
Prior to these changes, after connect action using any cloud drive connector, no information displayed to verify if the connection is established successfully or not.

After these changes, an information alert will be displayed after connect action to inform either the connection is success or failed.


azayati (approved):
LGTM

PR #1833, opened 2022-07-27
Task-58724: Delete old cloud drive connection buttons and content from attachments drawer

abenmbarek:
Displayed cloud drive connectors buttons from the attachment drawer code base should be removed.


apps/portlet-clouddrives/src/main/webapp/WEB-INF/gatein-resources.xml
```diff
@@ -176,7 +149,7 @@
 
   <portlet-skin>
     <application-name>clouddrives</application-name>
-    <portlet-name>CloudDrivePortlet</portlet-name>
+    <portlet-name>CloudDriveSettingsPortlet</portlet-name>
```
ahamdi, line 152: Do you need the whole styles ? doesn't it need some clean up ?

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
```diff
@@ -98,27 +98,6 @@
               class="uiIconLightGray uiIconAddFolder"
               @click="addNewFolder()"></a>
           </div>
-          <!-- Action buttons for extensionRegistry extensions -->
-          <div
-            v-for="action in attachmentsComposerActions"
```
ahamdi, line 103: IS this used by clouddrives or even other extensions ?
This should not be removed IMO, we will loose Attachement composer extensibility

ahamdi (changes requested):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
abenmbarek: For the moment, this block is used only to display the cloud drive connectors so i decided to remove it, if i keep it this block will be useless. Should i ?

apps/portlet-clouddrives/src/main/webapp/WEB-INF/gatein-resources.xml
abenmbarek: how ?

ahamdi: Check the CSS file and see if the styles are usefull for the drawer or not

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-drive-explorer/AttachmentsDriveExplorerDrawer.vue
ahamdi: We have to keep it, extensibility goes beyong Cloud drives, that was just one use case

apps/portlet-clouddrives/src/main/webapp/WEB-INF/gatein-resources.xml
abenmbarek: almost all the css classes are used in [clouddrive-dms.js](https://github.com/exoplatform/ecms/blob/feature/marketfit/apps/portlet-clouddrives/src/main/webapp/js/clouddrive-dms.js)

ahamdi (approved):

PR #1859, opened 2022-08-19
 TASK-59262: Several warning messages when uploading more than 20 files

akhanfir:
Prior to this change, when uplaod more than 21 files in the documents application, more than one warning message is displayed
To fix this, Only one warning message displayed when maximum number of attachments list equals 21


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentsUploadInput.vue
```diff
@@ -193,19 +193,21 @@ export default {
       }
 
       newAttachedFiles.filter(file => !this.attachments.some(f => f.title === file.title)).forEach((newFile,index) => {
-        this.queueUpload(newFile,index);
+        if (this.attachments.length === this.maxFilesCount) {
```
ahamdi, line 196: what happens for the case when the number of attachments is bigger than maxFileCount ? here you test just for the equality

ahamdi (approved):

PR #1865, opened 2022-08-29
Task-59140: Disconnect cloud drive account

IlhemEssaadi:
After these changes, when I disconnect  from my cloud drive account, the cloud drive folder will be deleted automatically and the connected drive account will also be removed from the settings and connector drawer.


azayati (approved):
LGTM

PR #1866, opened 2022-08-29
Task-59122: Add disconnect button for connected cloud drives

IlhemEssaadi:
After these changes, we are able to display a Disconnect button instead of Connect button when the connection to cloud drive is successfull.


azayati (approved):
LGTM

PR #1868, opened 2022-08-31
Task-58651_58652: Fix opening cloud drive documents from Google Clouddrive or one drive

IlhemEssaadi:
After these changes, we fixed the wrong display of opening documents from Google Clouddrive or one drive.


azayati (approved):
LGTM

PR #1870, opened 2022-09-01
TASK-57905: Document preview : document name displaying encoded items

hchaabene:
**prior to change,** when i upload a file the name of her contain special character **(%']:[|./)** replaced by number generated by a function **escapeIllegalJcrChars**() for example t'ask replaced by t_25ask
Fix : when i upload file i skipped that function  , to not replaced the name of file by number.


ahamdi (changes requested):
This should not be fixed like this : 

1. The code was not proposed on develop branch, which means the behavior between 6.4 and 6.3 will be different
1. Files will be saved on JCR , which means they won't be stored correctly when their names contain illegal chars from the list specified in JCR specification

Could you rework on the task considerig the remarks abve

PR #1875, opened 2022-09-07
Task-59577: Hide 3 dots button & edit button from Unified search results

abenmbarek:


core/search/src/main/java/org/exoplatform/services/wcm/search/connector/FileSearchServiceConnector.java
```diff
@@ -176,6 +176,18 @@ protected SearchResult buildHit(JSONObject jsonHit, SearchContext searchContext)
       ecmsSearchResult.setUrl(downloadUrl.toString());
     }
     ecmsSearchResult.setMetadatas(retrieveMetadataItems(id) );
+    ExtendedSession session2 = null;
```
azayati: you can avoid a new try/catch by adding a variable boolean isCloudDrive = node.hasProperty("ecd:driveUUID") in the above try/catch the set it directly here to the ecmsSearchResult

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #1889, opened 2022-09-29
fix: Fix previewing a shared document by its copied link - EXO-58927

Jihed525:
Prior to this change, we are not able to preview a shared document by its copied link since it is not possible to retrieve to corresponding symlink node using session.getNodeByUUID(). To fix that, we use extendedSession.getNodeByIdentifier in order to retrieve correctly the symlink node and preview it correctly.


azayati (approved):
LGTM

PR #1897, opened 2022-10-06
fix: Fix recreating the activity stream attachments folder after delete or move it from root path - EXO-59928

hakermi:
Prior to this change, when delete or move the activity stream attachments folder, the attachment drawer is no more able to recreate it again and upload attachments inside due to a bad manipulation while creating the folder, in backend service, and a missing internalization after recreation it in frontend in the attachment drawer. 
This PR should make sure to correctly create the folder with its correct name and node type and reinitialize the drawer in case of recreating the folder.


mkrout (approved):
LGTM

PR #1901, opened 2022-10-10
fix: Fix lost extension after renaming a document - EXO-60015

sofyenne:
Prior to change there is a regression caused by https://github.com/exoplatform/ecms/pull/1817, so the modification is reverted, no need to change the title since it is a property that does not need to be neither cleaned nor escaped for its illegal JCR characters.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
@@ -701,7 +701,7 @@ public class Utils {
 
   public static String cleanName(String oldName) {
     if (StringUtils.isEmpty(oldName)) return oldName;
-    String extension ="" ;
+    String extension = "" ;
     if(oldName.lastIndexOf(".") > -1){
       extension = oldName.substring(oldName.lastIndexOf("."));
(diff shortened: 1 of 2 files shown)


core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
```diff
@@ -701,9 +701,9 @@ public static String cleanString(String str) {
 
   public static String cleanName(String oldName) {
     if (StringUtils.isEmpty(oldName)) return oldName;
-    String extention ="" ;
+    String extension ="" ;
```
azayati:
```suggestion
    String extension = "" ;
```

azayati (changes requested):
see review

core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
sofyenne: done !

azayati (changes requested):
you need to change the source branch on feature/maintenance not develop

azayati (approved):
LGTM

PR #1919, opened 2022-10-20
fix: Fix lost extension after renaming a document - EXO-60015

sofyenne:
Prior to change there is a regression caused by #1817, so the modification is reverted, no need to change the title since it is a property that does not need to be neither cleaned nor escaped for its illegal JCR characters.


azayati (approved):
LGTM

PR #1920, opened 2022-10-24
feat: Resize enormous illustrations - EXO-60323

hakermi:
Prior to this change, image attachments are displayed and fetched with their original size which affects the perf on stream page and leading to a long load and render time. This PR should make sure to resize these images by using the jcr thumbnail service to create a custom thumbnails.


ahamdi (approved):

core/services/src/main/java/org/exoplatform/services/cms/impl/ImageUtils.java
```diff
@@ -73,9 +73,15 @@ public static InputStream scaleImage(BufferedImage image, int maxWidth, int maxH
...
-        thumbImage = Scalr.resize(image, Scalr.Method.BALANCED, maxWidth, maxHeight);      
+        int originWidth = image.getWidth();
+        int originHeight = image.getHeight();
+        Scalr.Mode fitMode = Scalr.Mode.AUTOMATIC;
+        if (originWidth > originHeight) {
+          fitMode = Scalr.Mode.FIT_TO_HEIGHT;
+        } else if (originHeight > originWidth) {
+          fitMode = Scalr.Mode.FIT_TO_WIDTH;
+        }
+        thumbImage = Scalr.resize(image, Scalr.Method.QUALITY, fitMode, maxWidth, maxHeight, Scalr.OP_ANTIALIAS);
```
rdenarie: Should we replace ImageUtils call by a call to ImageResizeService ?

rdenarie (changes requested):

core/services/src/main/java/org/exoplatform/services/cms/impl/ImageUtils.java
hakermi: Yes we can

rdenarie: We should do it in the same task so

hakermi: done

rdenarie: Can you rebase the PR ?

hakermi: rebased

rdenarie (approved):

PR #1922, opened 2022-10-25
fix: Fix recent documents display in my work tab of snapshot documents portlet - EXO-60243

akhanfir:
Prior to this change, my work tab of snapshot documents portlet is empty, no recent documents are displayed. It is caused by a 404 error mentioning that query param is required in rest API call. After this change, since query param is not mandatory, we have fixed that by removing the check on mandatory query pram in order to ensure displaying recent documents properly in my work tab.


azayati (approved):
LGTM

PR #1923, opened 2022-10-26
fix: fix owner property of activity stream attachments folder

Jihed525:
Prior to this change, when recreate the activity stream attachments folder, the session used was the one of the current user wich gives the owner permission to the current creator user. This PR should add an isSystem flag for create folder endpoint to manage such system folders with system session

(cherry picked from commit 705a4caf86b61753562ff908201f309f0e5108f4)


azayati (approved):
LGTM

PR #1924, opened 2022-10-26
fix: Fix source mode view problem on ckeditor when creating or editing a web content

Jihed525:
Prior to this change, we have many js errors in browser console and no action (save, save and close ...) is possible when we want to create or edit a web-content, use source mode view on ckeditor then return to normal mode view. This is due to a problem with suggester plugin which should be removed. After this commit, we ensure to remove this useless plugin in order to make other plugins work correctly.

(cherry picked from commit bbb8a54f71561254c93c5308ccf6190599bae0dc)


azayati (approved):
LGTM

PR #1928, opened 2022-11-03
fix: Fix documents disappears from "My work" snapshot documents widget tab when they are modified by another user - EXO-58370

IlhemEssaadi:
Prior to this change, when a user creates a document and the document is modified by another user, the document disappears from the list of tab "MY WORK". 
The problem is that the document appears only for the last modifier. After this change, we fix the Elasticsearch filter to display the documents for which I'm the author or the last modifier.


rdenarie (dismissed):
What if I'm not the author nor the last modifier ? 
As I worked on the document, I want it in My Last Documents

azayati (approved):
LGTM

PR #1930, opened 2022-11-04
feat: remove unneeded preload URLs - EXO-60649

GouadriaHanen:
prior to this change, we have warnings about preload problems,
after this change, no warning will be displayed as the preloaded URLs are removed from the dynamic container as they are no longer in use.


ahamdi:
@GouadriaHanen for what purpose was the portlet used ? could you find out why it was added on all pages ? Shouldn't we just remove it from the top container without removing completely its sources ?

GouadriaHanen:
> @GouadriaHanen for what purpose was the portlet used ? could you find out why it was added on all pages ? Shouldn't we just remove it from the top container without removing completely its sources ?

it is only injected into the dynamic container, it is not used anywhere else.

ahamdi (approved):

PR #1935, opened 2022-11-08
fix: Fix documents disappears from "My work" snapshot documents widget tab when they are modified by another user 

Jihed525:
Prior to this change, when a user creates a document and the document is modified by another user, the document disappears from the list of tab "MY WORK". The problem is that the document appears only for the last modifier. After this change, we fix the Elasticsearch filter to display the documents for which I'm the author or the last modifier.

(cherry picked from commit 4307c8ac50a7fac11737ccc3e0656186a88aa66c)


azayati (approved):
LGTM

PR #1942, opened 2022-11-14
fix: Image can't be saved after selection from computer -EXO-60730

akhanfir:
Prior to this change, when creating a new note and clicking on the upload image button from the insert plugin drawer of notes ckeditor, the upload process does not finish correctly. After this change the upload process works correctly for all browsers.


azayati (approved):
LGTM

PR #1949, opened 2022-11-23
fix: Fix personal drive shared document preview issues (breadcrumb, wrong folder displayed) - EXO-60838

GouadriaHanen:
Prior to this change, when retrieving a shared document, the original document was received, resulting in the wrong breadcrumb and path being displayed.
After this change, the retrieved document contains the breadcrumb and path of the shared document.

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
+++ b/core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
@@ -133,7 +133,7 @@ public class EntityBuilder {
     try {
       previewBreadcrumb = documentService.getFilePreviewBreadCrumb(attachmentNode);
     } catch (Exception e) {
-      LOG.error("Error while getting file preview breadcrumb " + originalAttachmentNode.getUUID(), e);
+      LOG.error("Error while getting file preview breadcrumb " + ((NodeImpl) originalAttachmentNode).getIdentifier(), e);
     }
     attachment.setPreviewBreadcrumb(previewBreadcrumb);
 
--- a/core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
+++ b/core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
@@ -25,6 +25,7 @@ import org.exoplatform.services.cms.link.LinkManager;
 import org.exoplatform.services.cms.mimetype.DMSMimeTypeResolver;
 import org.exoplatform.services.jcr.RepositoryService;
 import org.exoplatform.services.jcr.core.ExtendedSession;
+import org.exoplatform.services.jcr.impl.core.NodeImpl;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.services.wcm.utils.WCMCoreUtils;
@@ -95,7 +96,7 @@ public class EntityBuilder {
     }
 
     Attachment attachment = new Attachment();
-    attachment.setId(originalAttachmentNode.getUUID());
+    attachment.setId(((NodeImpl) originalAttachmentNode).getIdentifier());
     String attachmentsTitle = getStringProperty(originalAttachmentNode, "exo:title");
     attachment.setTitle(attachmentsTitle);
     String attachmentsPath = attachmentNode.getPath();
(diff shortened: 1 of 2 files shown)


core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
```diff
@@ -86,37 +86,37 @@ public static final Attachment fromAttachmentNode(RepositoryService repositorySe
...
+      originalDocumentNode = linkManager.getTarget(attachmentNode);
+      if (originalDocumentNode == null) {
         throw new ObjectNotFoundException("Target Node with of symlink " + attachmentId + " wasn't found");
       }
     }
 
     Attachment attachment = new Attachment();
-    attachment.setId(attachmentNode.getUUID());
-    String attachmentsTitle = getStringProperty(attachmentNode, "exo:title");
+    attachment.setId(originalDocumentNode.getUUID());
```
ahamdi: Just to make sure we use the same logic, why we use sometimes the UUID and other times the JCR identifier ?

GouadriaHanen: here for Attachment they use the UUID of the Node

azayati: Before this [fix commit](https://github.com/exoplatform/ecms/commit/e929b5b8ff82ab8b98af5ca13cee43316574551e) there were some exceptions related to the **getNodeByUUID** on a symlink since it is not referenceable, that's why we have changed in all related locations **getNodeByUUID**  by **getNodeByIdentitfier**. 
The problem now is that the displayed breacrumb corresponds to the original document not the symlink, so in the case of a user previewing the shared document (symlink) without permissions on the original document, the breacrumb will not be displayed correcly. IMO, in case of a symlink we should we should display its breacrumb and not the original document one.

core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
```diff
@@ -86,37 +86,37 @@ public static final Attachment fromAttachmentNode(RepositoryService repositorySe
...
+    String attachmentsTitle = getStringProperty(originalDocumentNode, "exo:title");
     attachment.setTitle(attachmentsTitle);
     String attachmentsPath = attachmentNode.getPath();
     attachment.setPath(attachmentsPath);
```
azayati, line 103: be careful about this lines if they should be modified also

azayati (changes requested):
see review

core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
```diff
@@ -86,37 +86,37 @@ public static final Attachment fromAttachmentNode(RepositoryService repositorySe
     } catch (ItemNotFoundException e) {
       throw new ObjectNotFoundException("Node with id " + attachmentId + " wasn't found");
     }
-
+    Node originalDocumentNode = attachmentNode;
```
azayati:
```suggestion
    Node originalAttachmentNode = attachmentNode;
```

azayati (dismissed):

core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
GouadriaHanen: we need the path of the shared attachment not the original one

core/services/src/main/java/org/exoplatform/services/attachments/utils/EntityBuilder.java
```diff
@@ -125,14 +126,14 @@ public static final Attachment fromAttachmentNode(RepositoryService repositorySe
...
-    String attachmentsVersion = getStringProperty(attachmentNode, "exo:baseVersion");
+    String attachmentsVersion = getStringProperty(originalAttachmentNode, "exo:baseVersion");
     attachment.setVersion(attachmentsVersion);
 
     LinkedHashMap<String, String> previewBreadcrumb = new LinkedHashMap<>();
     try {
       previewBreadcrumb = documentService.getFilePreviewBreadCrumb(attachmentNode);
     } catch (Exception e) {
-      LOG.error("Error while getting file preview breadcrumb " + attachmentNode.getUUID(), e);
+      LOG.error("Error while getting file preview breadcrumb " + originalAttachmentNode.getUUID(), e);
```
ahamdi:
```suggestion
      LOG.error("Error while getting file preview breadcrumb " + (NodeImpl) originalAttachmentNode).getIdentifier(), e);
```

azayati: @ahamdi I think that since the originalAttachmentNode will be original node not the symlink, it is always referencable so getUUID will be also good IMO

ahamdi (approved):

PR #1952, opened 2022-11-28
feat: Add an option to allow directly open attachments in editor by bypassing the preview step - EXO-60484

hakermi:
…
Add an option to allow directly open attachments in editor by bypassing the preview step in attachments list drawer


mkrout (approved):
LGTM

PR #1958, opened 2022-12-15
fix: Refresh needed to display created document - EXO-61166

GouadriaHanen:
prior to this change, a refresh is needed after creating to display into the files list.
after this change, after adding the document  an event is sent to update the list and success alert is displayed

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
@@ -179,8 +179,6 @@ export default {
         doc.drive = this.currentDrive.title;
         doc.date = doc.created;
         this.$root.$emit('add-new-created-document', doc);
-        this.$root.$emit('end-loading-attachment-drawer');
-        document.dispatchEvent(new CustomEvent('entity-attachments-updated'));
         this.$root.$emit('attachments-notification-alert', {
           message: this.$t('attachments.upload.success'),
           type: 'success',
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
@@ -180,6 +180,11 @@ export default {
         doc.date = doc.created;
         this.$root.$emit('add-new-created-document', doc);
         this.$root.$emit('end-loading-attachment-drawer');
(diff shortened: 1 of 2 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
```diff
@@ -180,6 +180,11 @@ export default {
         doc.date = doc.created;
         this.$root.$emit('add-new-created-document', doc);
         this.$root.$emit('end-loading-attachment-drawer');
+        document.dispatchEvent(new CustomEvent('entity-attachments-updated'));
```
ahamdi: There is already an event above ```this.$root.$emit('add-new-created-document', doc);```
Isn't it sufficient ? maybe we need to look at it and see why it does not update the list of attachments ? WDYT ?

ahamdi (changes requested):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
GouadriaHanen: `this.$root.$emit('add-new-created-document', doc) :`  it is used to push the created file to the list of files in the drawer and not in the drives files list  in the drives app.
we need to update the files list in the drives app

ahamdi: shouldn't we then include the global event **entity-attachments-updated** inside the other event to make sure it is fired whenever a document is added ?

GouadriaHanen: done!

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachment-document-creator/AttachmentCreateDocumentInput.vue
```diff
@@ -179,7 +179,11 @@ export default {
         doc.drive = this.currentDrive.title;
         doc.date = doc.created;
         this.$root.$emit('add-new-created-document', doc);
-        this.$root.$emit('end-loading-attachment-drawer');
+        document.dispatchEvent(new CustomEvent('entity-attachments-updated'));
```
ahamdi: @GouadriaHanen could you remove the instruction from here ?

GouadriaHanen: done

ahamdi (approved):

PR #1980, opened 2023-01-12
feat: Remove administration application registry category and configure tools application registry category - EXO-61480

sofyenne:
This change is going to add the document portlet under the tools application registry category and remove the administration application registry category


azayati (approved):
LGTM

PR #1984, opened 2023-01-16
fix: Upload and create document error - EXO-61814

GouadriaHanen:
prior to this change, it is not possible to upload or create a document  since the current drive and the DefaultDestinationFolderPath are not initialized
after this change, the attributes are initialized, and adding documents works perfectly


ahamdi (commented):
Could you instead check why the event `attachments-default-folder-path-initialized` is not working in this case ?

GouadriaHanen:
> attachments-default-folder-path-initialized
this removed part, [code](https://github.com/exoplatform/ecms/commit/34e65de4b40769b9a515d6701facb3da4b7525f5#diff-6ac38e632561521e5973749469d5eb9f0c68f1376235c97d30c3e8e43f9b6c5fL515-L528) , is responsible for broadcasting the event,  only in the case of an entityId the event could be broadcast which is not the case here.

ahamdi (approved):

PR #1996, opened 2023-02-07
fix: on document preview, with a large file, the click on download button do nothing - EXO-62303

rdenarie:
Before this fix, when the user clicks on download button on document preview, the item was fetch by javascript. In case of large item, it could be very long to execute This fix modify the behaviour, so that the user directly download the item. there is no more javascript fetch

### DIFF
--- a/apps/resources-wcm/src/main/webapp/javascript/eXo/ecm/document-preview.js
+++ b/apps/resources-wcm/src/main/webapp/javascript/eXo/ecm/document-preview.js
@@ -372,15 +372,14 @@
         }
 
         if (!this.settings.doc.size == "" && !this.isDownloadStatusActivated) {
-            if (this.settings.doc.downloadUrl.includes('javascript:')) {
-              html += '      <li><div class="downloadBtn">' +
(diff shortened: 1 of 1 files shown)


ahamdi (commented):
@rdenarie Why there were two ways to download document ? isn't there any risk by disabling it ?
It may cause regression for this task https://community.exoplatform.com/portal/dw/tasks/taskDetail/46540

rdenarie:
> @rdenarie Why there were two ways to download document ? isn't there any risk by disabling it ? It may cause regression for this task https://community.exoplatform.com/portal/dw/tasks/taskDetail/46540

What do you mean but "there were 2 ways to download document" ? Do you mean the if (this.settings.doc.downloadUrl.includes('javascript:')) { ? Because I kept this part, I simply have factorized the code

About the regression you link : 
Islem say in her last comment : 
"when I open an image in the preview then I click on the three dots and I select "Download", the image is not downloadable & it will be displayed on fullscreen" : when I look at the fix done by Boubaker, he made it to be able to really dowload the image when clicking on "Download", and not display it in the browser. 
I replace his fix by the attribute html download, which force the download of the content, even if it is an image.

ahamdi:
Ok thank you foe the details

ahamdi (approved):

PR #1998, opened 2023-02-10
fix : Decide whether to keep or create a new version of the file when uploading a file with an accented name - EXO-62290

sofyenne:
Prior to this change, if we wanted to upload a previously uploaded file with an accented name, the file would be uploaded without the option to either keep or create a new version. The problem was that the node name was cleaned using the method "cleanNameWithAccents" and saved without accents. As a result, if we wanted to check for its existence, we would have to clean the name as well. This change addresses this issue.

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java
+++ b/core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java
@@ -305,7 +305,6 @@ public class FileUploadHandler {
     DocumentBuilder builder = factory.newDocumentBuilder();
     Document fileExistence = builder.newDocument();
     fileName = Utils.cleanNameWithAccents(fileName);
-    fileName = Text.escapeIllegalJcrChars(fileName);
     fileName = fileName.replaceAll(FILE_DECODE_REGEX, "%25");
     fileName = URLDecoder.decode(fileName,"UTF-8");
     fileName = fileName.replaceAll(FILE_DECODE_REGEX, "-");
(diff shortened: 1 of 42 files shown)


core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java
```diff
@@ -304,6 +304,7 @@ public Response checkExistence(Node parent, String fileName) throws Exception {
     DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
     DocumentBuilder builder = factory.newDocumentBuilder();
     Document fileExistence = builder.newDocument();
+    fileName = Utils.cleanNameWithAccents(fileName);
     fileName = Text.escapeIllegalJcrChars(fileName);
```
ahamdi: Utils.cleanNameWithAccents calls already Text.escapeIllegalJcrChars (from JCR module) , is it needed to call Text.escapeIllegalJcrChars (gtom ECMS module) here ?
Side note, I don't get why we have two implementations of Text.escapeIllegalJcrChars in  our source code !

ahamdi (changes requested):

core/connector/src/main/java/org/exoplatform/wcm/connector/FileUploadHandler.java
sofyenne: sorry i wasn't paying attention to the call of the `Text.escapeIllegalJcrChars` method by the `Utils.cleanNameWithAccents` method , i was add just the messing `Utils.cleanNameWithAccents` call .
Call of unnecessary method removed !

ahamdi (approved):

PR #2001, opened 2023-02-15
fix: Error when create new resource storage dw - EXO-62334

GouadriaHanen:
before this change, when initializing portal configurations, an error is detected when creating a new DW resource storage because its portal storage "/sites" is not yet created 
after this modification, we should avoid creating a new resource storage when its portal storage is not yet created

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
@@ -145,7 +145,12 @@ public class LivePortalManagerServiceImpl implements LivePortalManagerService, S
     if (StringUtils.equals(portalConfigService.getGlobalPortal(), portalName)) {
       return null;
     }
-    Node portalsStorage = getLivePortalsStorage(sessionProvider);
+    Node portalsStorage = null ;
+    try {
+      portalsStorage = getLivePortalsStorage(sessionProvider);
+    } catch (PathNotFoundException e) {
+      return null;
+    }
     if (portalsStorage != null) {
       return portalsStorage.getNode(portalName);
     }
--- a/core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
+++ b/core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
@@ -16,10 +16,16 @@
  */
 package org.exoplatform.services.wcm.portal.impl;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-import javax.jcr.*;
+import javax.jcr.Node;
+import javax.jcr.NodeIterator;
+import javax.jcr.PathNotFoundException;
+import javax.jcr.Session;
 import javax.jcr.query.Query;
 import javax.jcr.query.QueryResult;
 
@@ -29,12 +35,9 @@ import org.exoplatform.portal.config.DataStorage;
 import org.exoplatform.portal.config.UserPortalConfigService;
 import org.exoplatform.portal.config.model.PortalConfig;
 import org.exoplatform.services.jcr.RepositoryService;
-import org.exoplatform.services.jcr.access.PermissionType;
 import org.exoplatform.services.jcr.core.ExtendedNode;
 import org.exoplatform.services.jcr.core.ManageableRepository;
 import org.exoplatform.services.jcr.ext.common.SessionProvider;
-import org.exoplatform.services.jcr.ext.distribution.DataDistributionManager;
-import org.exoplatform.services.jcr.ext.distribution.DataDistributionMode;
 import org.exoplatform.services.listener.ListenerService;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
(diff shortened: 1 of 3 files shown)


core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
```diff
@@ -145,7 +145,12 @@ public final Node getLivePortal(final SessionProvider sessionProvider,
     if (StringUtils.equals(portalConfigService.getGlobalPortal(), portalName)) {
       return null;
     }
-    Node portalsStorage = getLivePortalsStorage(sessionProvider);
+    Node portalsStorage = null ;
```
ahamdi: Could you instead force initiating the service LivePortalManagerServiceImpl to start after the creation of needed JCR folder ?

ahamdi (changes requested):

core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
GouadriaHanen: it is not possible even with fake dependency

ahamdi: I checked and you are right. What happens if we let the LivePortalManager create the /sites folder if it does not exist?

ahamdi (commented):
What happens if we let the LivePortalManager creates the /sites folder if it does not exist?

core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
```diff
@@ -193,7 +197,14 @@ private Node getLivePortalsStorage(final SessionProvider sessionProvider) throws
     String portalsStoragePath = locationEntry.getPath();
     ManageableRepository manageableRepository = repositoryService.getCurrentRepository();
     Session session = sessionProvider.getSession(workspace, manageableRepository);
-    return (Node)session.getItem(portalsStoragePath);
+    try {
+      return (Node) session.getItem(portalsStoragePath);
+    } catch (PathNotFoundException e) {
+      if (SITES_PATH.equals(portalsStoragePath)) {
```
ahamdi: What about starting the NodeHierarchyCreator here ?  could it solve the issue ?

GouadriaHanen: it is not possible!

core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
```diff
@@ -345,4 +356,23 @@ private void checkAllPortalSitesCreated() throws Exception {
...
+    permissions.add(PermissionType.SET_PROPERTY);
+    permissions.add(PermissionType.REMOVE);
+    List<String> readPermissions = new ArrayList<String>();
+    readPermissions.add(PermissionType.READ);
+    permissionsMap.put("*:/platform/web-contributors", permissions.toArray(new String[permissions.size()]));
+    permissionsMap.put("*:/platform/administrators", permissions.toArray(new String[permissions.size()]));
+    permissionsMap.put("any", readPermissions.toArray(new String[readPermissions.size()]));
+
+    return dataDistributionManager.getDataDistributionType(DataDistributionMode.NONE)
+                                   .getOrCreateDataNode(rootNode, SITES_PATH, null, null, permissionsMap);
```
rdenarie: today the sites node is created by configuration. Your reproduce the same configuration here, but, if tomorrow the configuration change, we will have a problem. 

I dont think we should create the sites node like this if it not exists, but ensure the services to start in the correct order.

WDYT @ahamdi ?

core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
rdenarie: Why it is not possible ?

core/services/src/main/java/org/exoplatform/services/wcm/portal/impl/LivePortalManagerServiceImpl.java
ahamdi: @rdenarie I agree, we need to crate the node using configuration, that is why I asked to check the NodeHierarchyCreator

core/services/src/main/java/org/exoplatform/services/wcm/portal/listener/CreateLivePortalEventListener.java
```diff
@@ -178,4 +184,18 @@ public UserPortalConfigService getPortalConfigService() {
...
+    if (!isNodeHierarchyCreatorInitialized) {
+      LOG.info("init nodeHierarchyCreator ...");
+      try {
+        this.nodeHierarchyCreator.init();
+        isNodeHierarchyCreatorInitialized = true;
+        LOG.info("nodeHierarchyCreator initialized");
+
+      } catch (Exception e) {
+        LOG.error("error when initializing nodeHierarchyCreator", e);
+      }
```
rdenarie: Do we need to init the nodeHierarchyCreator ? 
As it is passed as parameter, it should be started and initialized when we arrive to this point ?

rdenarie (dismissed):

core/services/src/main/java/org/exoplatform/services/wcm/portal/listener/CreateLivePortalEventListener.java
GouadriaHanen: the CreateLivePortalEventListener starts after the creation of portals at the gatein-portal level with a Startable service that has priority to start before other services so we need to force the start of the nodeHierarchyCreator service to create the "/sites"

ahamdi (approved):

PR #2038, opened 2023-03-14
fix : Adding documents and folders in a read only shared folder should not be possible - EXO-63084

sofyenne:
This change is going to disallow the ability to add a new folder inside a shared folder with read only permission


ahamdi (approved):

exo-swf:
Your PR triggers too many exo-ci builds! Please finish your work and then, set your PR ready! Thank you

PR #2045, opened 2023-03-16
fix: Document warning snack bar during document upload conflict is not in the foreground - EXO-62942

IlhemEssaadi:
After this fix, the snack bar will be displayed in the foreground in front of the grey overlay by removing absolute property in order to get the right z-index value.

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/snackbar/AttachmentsNotificationAlerts.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/snackbar/AttachmentsNotificationAlerts.vue
@@ -4,7 +4,6 @@
     color="transparent"
     elevation="0"
     app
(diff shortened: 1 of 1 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/Attachment.vue
```diff
@@ -19,7 +19,8 @@
         :supported-documents="supportedDocuments"
         :attachments="attachments"
         :open-attachments-in-editor="openAttachmentsInEditor" />
-      <attachments-notification-alerts />
+      <attachments-notification-alerts
```
ahamdi: Could you check if using the shared component `<exo-notification-alert />` solves the issue without adding the new CSS class ?

ahamdi (approved):

PR #2054, opened 2023-03-23
fix: Fix change location option issue when uploading file - EXO-62498

IlhemEssaadi:
Prior to this change, when we upload a file and change the location option, all accessible drives are displayed.
After this fix, if I have not changed the location of any of the uploaded documents, the current location is proposed to change the location of the uploaded file, otherwise the location where the upload was initially done is proposed.


ahamdi (commented):
Could you add a description to this PR and provide details about the problem causes and how the fix was made ?

ahamdi (approved):

PR #2065, opened 2023-03-27
fix: Replace & char with underscore in file technical name - EXO-62912 (#2058)

Jihed525:
Prior to this change, when create a file will commercial & the char remains the same in the technical name. This PR makes sure to replace the char with undersoce and so prevent any future issue caused by the sepcial char

(cherry picked from commit 727284bcca866699af057d7f59e1684b435395a0)


hakermi (approved):

sofyenne (commented):
@Jihed525 Task id is wrong  !!

sofyenne (approved):

PR #2073, opened 2023-04-03
fix : Add new document version using only office generate no content activity -EXO-63508

sofyenne:
Prior to this change when we create a new document version using only office an empty activity is generated , the problem was a regression caused by this commit https://github.com/exoplatform/ecms/commit/bc309e1725375179f77f102fd52c42704f8eceb9
This change addresses this issue


ahamdi (approved):

ahamdi:
@sofyenne @Jihed525 shouldn't we have an upgrade plugin for this fix ?

PR #2085, opened 2023-05-05
fix : Do not allow uploading more than the defined maximum number of documents - EXO-61841

sofyenne:
Before this change, we were able to upload more documents than the defined maximum limit.
The problem was the incorrect check for the uploaded file number.
With this change, it will not be possible to upload more than the defined maximum number.

### DIFF
--- a/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentsUploadInput.vue
+++ b/apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentsUploadInput.vue
@@ -199,15 +199,16 @@ export default {
         });
       }
 
-      newAttachedFiles.filter(file => !this.attachments.some(f => f.title === file.title)).forEach((newFile, index) => {
-        if (index === this.maxFilesCount || this.maxFilesCount === 0) {
+      newAttachedFiles.filter(file => !this.attachments.some(f => f.title === file.title)).every((newFile, index) => {
+        if (index === 3 || this.maxFilesCount === 0) {
           this.$root.$emit('attachments-notification-alert', {
             message: this.maxFileCountErrorLabel,
             type: 'error',
           });
-          return;
+          return false;
         } else {
           this.queueUpload(newFile);
+          return true;
         }
       });
       this.$refs.uploadInput.value = null;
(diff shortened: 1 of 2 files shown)


apps/portlet-documents/src/main/webapp/vue-app/attachment/components/snackbar/AttachmentsNotificationAlerts.vue
```diff
@@ -45,7 +45,9 @@ export default {
   },
   methods: {
     addAlert(alert) {
-      if (alert) {
+      //check if the alert is already displayed
+      const alertExists = this.alerts.some((existingAlert) => existingAlert.message === alert.message && existingAlert.type === alert.type);
```
ahamdi: Will this disable also alerts for successful file uploads too for example ?

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentsUploadInput.vue
```diff
@@ -200,13 +200,11 @@ export default {
       }
 
       newAttachedFiles.filter(file => !this.attachments.some(f => f.title === file.title)).forEach((newFile, index) => {
```
ahamdi: If you need to break the loop , you can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every instead of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

ahamdi (changes requested):

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/attachments-upload-components/AttachmentsUploadInput.vue
sofyenne: done , Thanks for this helpful information , I was already searching for it  !

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/snackbar/AttachmentsNotificationAlerts.vue
sofyenne: Yes this check to no display any attachment alert already displayed in this moment !

ahamdi: could you check if this is expected already ? I believe we need to have as many alerts as needed when uploading files .

exo-swf:
Your PR triggers too many exo-ci builds! Please finish your work and then, set your PR ready! Thank you

apps/portlet-documents/src/main/webapp/vue-app/attachment/components/snackbar/AttachmentsNotificationAlerts.vue
sofyenne: Done !

ahamdi (approved):

PR #2097, opened 2023-05-15
feat: Add trashcleaner job - EXO-63795

rdenarie:
This feature create a job to regularly clear the trash
Two properties are used :
exo.trashcleaner.cron.expression=0 0 20 ? * SAT
The job run each saturday at 8PM
exo.trashcleaner.lifetime=30
The job will delete all content which are in trash since more than 30 days


core/services/src/main/java/org/exoplatform/services/cms/documents/job/TrashCleanerJob.java
```diff
@@ -0,0 +1,226 @@
+package org.exoplatform.services.cms.documents.job;
```
ahamdi, line 18: I didn't mention this in the maintenance PR, shouldn't we have a header for this added file  ?

rdenarie: Yes we should. I add it

ahamdi (approved):

PR #2103, opened 2023-05-15
fix : preventing users from re-uploading documents to the same attachment drawer after reaching the maximum limit - EXO-64261

sofyenne:
This fix is going to prevent users from re-uploading documents to the same attachment drawer after reaching the maximum limit


ahamdi (changes requested):
@sofyenne Could you fix the conflicts ?

sofyenne:
> @sofyenne Could you fix the conflicts ?

Done !

ahamdi (approved):

PR #2106, opened 2023-05-17
fix: Enable to upload a file containing an '+' character in its name - EXO-64192

sofyenne:
Before this change, when we uploaded a file with a name containing the '+' character, we were unable to preview or download it. The problem was that the '+' character in the file's path was being replaced by a space. With this change, we will add the '+' character to the special characters list, and it will be replaced by an underscore character in the node path. This modification will resolve the issue.


core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
```diff
@@ -736,7 +736,7 @@ public static String cleanName(String oldName) {
     if (StringUtils.isEmpty(oldName)) {
       return oldName;
     }
-    return replaceSpecialChars(oldName, "&#*?!@.'\"\t\r\n$\\><:;[]/|");
+    return replaceSpecialChars(oldName, "&#*?!@.'\"\t\r\n$\\><:;[]/|+");
```
ahamdi, line 739: The character '+' is not among illegal JCR chars, so users are allowed to use it in file names.
Could you find another way to fix the issue without replacing '+' characters

ahamdi (changes requested):

PR #2133, opened 2023-06-15
fix: Gif without animation in news illustration EXO-64163.

akhanfir:
Prior to this change, when create news in space whoose illustration image is a gif and post it, the gif is without animation. To fix this problem, when the image is of type gif, returned this image without doing the resize. After this change, illustration image is animated as it is the case in edit mode.

### DIFF
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <artifactId>ecms</artifactId>
     <groupId>org.exoplatform.ecms</groupId>
-    <version>6.5.x-SNAPSHOT</version>
+    <version>6.5.x-maintenance-SNAPSHOT</version>
   </parent>
   <artifactId>ecms-apps</artifactId>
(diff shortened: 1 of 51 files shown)


rdenarie (changes requested):
I'm afraid that your solution is not complete.
You only treat the part related to news, as it is described in the task.

But we should also fix it for other element which call ImageResizeService like avatar, banner, activities.
I think you should make the same fix in ImageThumbnailService, and update the expected behaviour for the task.

cc @ahamdi

ahamdi:
> 

@rdenarie The problem is fixed for activities and News. It fixes all usage of the ImageResizeService for contents uploaded in JCR
But it won't fix the avatars/banners as their processing changed on Meeds. The images are now cropped and we can no more get a Gif once it is uploaded and processed by the cropping tool.

rdenarie (approved):
Yes your right, so it is ok for me

PR #2175, opened 2023-09-20
fix : Enable to create a news with title contains a '.' character followed by a special character - EXO-66277

sofyenne:
Before this change we were unable to create a news with a title containing a '.' character followed by another special character , this issue was due to the clean name method which used the '.' to identify the file's extension and did not clean the last part of the name .
This fix will update the clean name method to check the node type before cleaning the node name .

### DIFF
--- a/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
+++ b/core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
@@ -705,8 +705,11 @@ public class Utils {
   }
   
   private static String replaceSpecialChars(String name, String specialChars) {
+    return replaceSpecialChars(name, specialChars, NodetypeConstant.NT_FILE);
+  }
+  public static String replaceSpecialChars(String name, String specialChars, String nodeType) {
     String extension = "";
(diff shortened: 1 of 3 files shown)


core/services/src/main/java/org/exoplatform/services/cms/impl/Utils.java
```diff
@@ -705,8 +705,11 @@ public static String cleanDocumentTitle(String title) {
   }
   
   private static String replaceSpecialChars(String name, String specialChars) {
+    return replaceSpecialChars(name, specialChars, NodetypeConstant.NT_FILE);
+  }
+  public static String replaceSpecialChars(String name, String specialChars, String nodeType) {
     String extension = "";
-    if (name.lastIndexOf(".") != -1) {
+    if (nodeType.equals(NodetypeConstant.NT_FILE) && name.lastIndexOf(".") != -1) {
```
ahamdi, line 712:
```suggestion
    if (NodetypeConstant.NT_FILE.equals(nodeType) && name.lastIndexOf(".") != -1) {
```

ahamdi (approved):

PR #2214, opened 2023-11-28
fix: Update the ecms applications images of page editor - EXO-67096

IlhemEssaadi:
After this change, we update the default images for each application in ecms.


azayati (approved):
LGTM

PR #2245, opened 2023-12-12
fix: Remove unneeded editor resizing after uploading an image - EXO-67793

ahamdi:
After copying pasting an image, the content of the note/news is hidden and we have a blank sheet.
The fix removes an unneeded resizing instruction of the editor that was setting the height to 0.


azayati (approved):
LGTM

PR #2246, opened 2023-12-13
fix: Apply the border radius for the transfer rules application - EXO-67638

sofyenne:
Before this modification, no border radius was applied to the transfer rules application. This change will now implement the border radius on the transfer rules application


azayati (approved):
LGTM

PR #2252, opened 2023-12-20
fix: Dissallow image inserting from desktop and pasting in the single note view application - EXO-68249

sofyenne:
Prior to this change we were able to insert/paste images from the desktop in the single note view application , this change is going to disallow  image inserting/pasting from the desktop in the single note view application.


ahamdi (commented):
Could you find a way to extract this logic outside of the plugin and enable/disable it based on the type of application ?

ahamdi (approved):

PR #2253, opened 2023-12-22
Include missing icu4j.jar in ecms packaging after upgrade to tomcat 10

azayati:


rdenarie (changes requested):
can you try to update the version of icu library in maven-exo-depmgt-pom ? 
we can check with the last one 74.2

rdenarie (approved):

PR #2262, opened 2023-12-27
fix: fix Error message and endless progress bar in drawer when adding a new version to a file - EXO-67382

GouadriaHanen:
Before this change, When creating a new version of a file the  controlUpload emit too many times events to re-upload the same file which had been removed from the upload handler from the first time
After this change, the  controlUpload emits the event when the file is not in process


azayati (approved):
LGTM

PR #2267, opened 2024-01-04
fix: fix upload progress and endless progress bar in drawer when adding a new version to a file - EXO-67382

GouadriaHanen:
before this change, while controlling upload and before finishing the the set timeout, the uploadId is cleared after finishing the upload causing retrieve the uploadProgress for null uploadId which has 0 as progress percent causing upload not finished popup
after this change, the set of the time out to control the upload of the upload progress only when the uploadId is not null


mkrout (approved):
LGTM

PR #2268, opened 2024-01-04
fix: updating a post creates an activity folder if ASD is deleted - EXO-68609

mkrout:
Prior to this fix, attachments of activities were stored under the ASD folder in case of creation and under a folder named with the activity ID in case of edit, and in some cases the asd folder (or Entity folder) are duplicated
This fix ensures that documents are added to the right folder and avoids folder duplication for any entity type. For activities, all documents are stored under the ASD folder in edit or creation cases.


GouadriaHanen (approved):
LGTM

akhanfir (approved):

PR #2275, opened 2024-01-10
fix : [AD Sync] After modification of a property, old value are regularly displayed in user profil - EXO-68774 - meeds/meeds-io#1533

rdenarie:
Before this fix, when the platform was run with eXo 6.4.3, custom ad attributes synchronized are store in gatein profile and in social profile In 6.4.4, the sync change, and attributes are directly stored in social profile. But, when the value exists in gatein profile, we can observe this problem : If attribute change in AD, the synchronization put it in social profile only. The correct new value is visible in user profile
When the user logs in, the social profile is synced, then the gatein profile is synced. As the gatein profile contains old values, the old values are pushed in the social profile This fix remove the listener SocialUserProfileEventListener, which is no more necessary as all ad attribute are directly pushed in social profile.


mkrout (approved):
LGTM

PR #2297, opened 2024-01-30
fix: fix Shortcut of folder content not displayed from backoffice - EXO-69082

GouadriaHanen:
Before this change, In the back office when opening a shortcut folder a PathNotFound exception was thrown since it could not find the shortcut element with the shortcut path causing a non-display of the shortcut elements
After this change, when a node is not found with a given path, we get the item from the shortcut target node


mkrout (approved):
Lgtm

PR #2322, opened 2024-03-08
fix: Display file size including version size in DocumentsSizeGadget -  EXO-70223

rdenarie:
Before this modification, the document size gadget compute size without take care of file versions This change add a field in file index to store fileSizeWithVersion


azayati (approved):
LGTM

PR #2325, opened 2024-03-12
fix: Activity stream documents folder creation issue - EXO-68612 (#2321)

mkrout:
Before this change, "System" folder created by the attachement drawer for entities like task, activity, process was created by the user session, and having the same permissions of other folders, so any user member of the space have the ability to rename, delete, or move this folders that can lead to loose attachements.
The proper way to fix this issue is to set the appropriate permissions to avoid this problem, but this cannot be possible with what we have as access management in the document app (we have only edit and view permissions).
As a temporary solution in this commit, all this folders will be created by a system session, so in the document app level we can recognize them and don't allow simple users to edit/move/delete them


sofyenne (approved):

Jihed525 (approved):

azayati (approved):
LGTM

PR #2355, opened 2024-05-16
fix: Create an extension for content attachments display - EXO-71746 - Meeds-io/meeds#1962

hakermi:
Create an extension for content attachments display


azayati (approved):
LGTM

PR #2359, opened 2024-05-22
feat: Add a share attachments listener to be used in content - EXO-71115 - Meeds-io/MIPs#130

hakermi:
Add a share attachments listener to be used in content


azayati (approved):
LGTM

PR #2364, opened 2024-06-10
fix: Fix resizing images and alignment via the selectImage dialog and toolbar options - EXO-70243_EXO-72349

hakermi:
Fix resizing images and alignment via the selectImage dialog and toolbar options


azayati:
@hakermi no need to meeds issue here since it is eXo codebase (ECMS) not meeds

azayati (approved):
LGTM

PR #2365, opened 2024-06-12
fix: Fix resizing images and alignment via the selectImage dialog and toolbar options - EXO-70243,EXO-72349

hakermi:
Fix resizing images and alignment via the selectImage dialog and toolbar options


azayati (approved):
LGTM

PR #2366, opened 2024-06-12
fix: Image properties context menu option appears even if the selection is not an image - EXO-72349

hakermi:
Image properties context menu option appears even if the selection is not an image


azayati (approved):
LGTM

PR #2387, opened 2024-09-18
fix : Remove unused function no more available in portal module

rdenarie:


azayati (approved):
LGTM

hbenali:
Could you modify the PR title (instead of external commit hash use https://github.com/Meeds-io/gatein-portal/pull/970) ? thanks

PR #2420, opened 2024-10-21
feat: Update article body image permissions handling - EXO-74441 - Meeds-io/MIPs#161

sofyenne:


core/services/src/main/java/org/exoplatform/services/attachments/listener/UpdateContentPermissionsListener.java
```diff
@@ -61,24 +63,33 @@ public class UpdateContentPermissionsListener extends Listener<Map<String, Objec
...
+    this.attachmentStorage = attachmentStorage;
   }
 
   @Override
   public void onEvent(Event event) throws Exception {
+
+    // set system session provider to avoid SessionProviderService NullPointerException with spring context
+    SessionProvider sessionProvider = SessionProvider.createSystemProvider();
+    SessionProviderService sessionProviderService = ExoContainerContext.getService(SessionProviderService.class);
+    sessionProviderService.setSessionProvider(null, sessionProvider);
```
azayati, line 77: - you can put this directly into the updateNodePermissions method
- why not keeping SessionProviderService injected in the constructor ? it already works well before

azayati (changes requested):

core/services/src/main/java/org/exoplatform/services/attachments/listener/UpdateContentPermissionsListener.java
sofyenne: Because it is always null when injected by the constructor, it is annotated with @DefinitionByType so it is not injected when using the Spring context. Additionally, we cannot include this part of code directly in the updateNodePermissions method due to the dependency of the SessionProviderService on the attachment storage!

azayati (approved):
LGTM

PR #2529, opened 2025-12-11
Fix: Special chars are converted to underscores when files are uploaded one by one - EXO-83146

akhanfir:
Before this change, go to doc app of spacex and upload files whose title contain special chars, all the mentioned chars are replaced by underscores. To fix this problem, remove the clean from title when saving file. After this change, the title is displayed with these tanks.

### DIFF
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <artifactId>ecms</artifactId>
     <groupId>org.exoplatform.ecms</groupId>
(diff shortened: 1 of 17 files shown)


rdenarie (changes requested):
The modification you propose has been added to fix this task https://community.exoplatform.com/portal/dw/tasks/taskDetail/61256

So, if we remove it, we loose the fix of the initial task.
So please review the content of the initial task, and the current task, and then ask to someone in the functionnal team what is the real expected

PR #2550, opened 2026-04-23
fix: Document content anonymously accessible by jcr id - EXO-86312

mkrout:
Prior to this commit, using the "/portal/rest/pdfviewer/repository/collaboration/{uuid}" return the requests file even if the users don't have access rights.

This commit fix this issue by using the user session to retrieve the document.

### DIFF
--- a/core/connector/src/main/java/org/exoplatform/wcm/connector/viewer/PDFViewerRESTService.java
+++ b/core/connector/src/main/java/org/exoplatform/wcm/connector/viewer/PDFViewerRESTService.java
@@ -28,11 +28,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.imageio.ImageIO;
+import javax.jcr.AccessDeniedException;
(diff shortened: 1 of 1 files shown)


Jihed525 (approved):

core/connector/src/main/java/org/exoplatform/wcm/connector/viewer/PDFViewerRESTService.java
```diff
@@ -130,20 +140,41 @@ public Response getCoverImage(@PathParam("repoName") String repoName,
...
+          return Response.status(Response.Status.NOT_FOUND).entity("Requested document not found").build();
+        }
+        ExtendedNode extendedNode = (ExtendedNode) currentNode;
+        List<AccessControlEntry> permsList = extendedNode.getACL().getPermissionEntries();
+        boolean isPublic = false;
+        for (AccessControlEntry accessControlEntry : permsList) {
+          if (StringUtils.equals(IdentityConstants.ANY, accessControlEntry.getIdentity())) {
+            isPublic = true;
+            break;
+          }
```
rdenarie, line 166: Does this part is necessary  ?
if user is anonym, should we create a anonymous session ? and as there is a "any" permission, it can access to the document ?

PR #2554, opened 2026-06-12
fix: prevent creation of default personal drive folders during user creation - EXO-87032

sofyenne:
Prior to this change, some default personal drive folders (Videos, Music, Search, etc.) were automatically created when a new user was created. This change prevents the creation of these default personal drive folders during user creation.


ecm-wcm-extension/src/main/webapp/WEB-INF/conf/dms-extension/jcr/jcr-configuration.xml
```diff
@@ -145,195 +145,8 @@
                     </field>
                   </object>
                 </value>
-                <value>
```
boubaker, line 148: Please pay attention to places where it's used

Image

boubaker:
IMO we make the same thing as https://github.com/exoplatform/data-upgrade/pull/292/changes by adding the `exo;hiddenable` mixin type, WDYT ?

boubaker (dismissed):

boubaker (dismissed):

boubaker (approved):

boubaker:
please make sure that Documents Build and Favorite & Share Features are ok as well

